PR #1, opened 2015-05-05
Update .gitignore

lochel:
We need a proper .gitignore file. This can be used as a starting point.


lochel:
There is much more that need to be included in the .gitignore file.

PR #17, opened 2015-06-05
Some changes to "remove simple equations" and related stuff

lochel:
* [fixed] propagation of variable attribute "unreplaceable" within the back end
* [fixed] all variables that have unreplaceable=true will now get skipped by "removeSimpleEquations"
* [improved] reimplemented function selfGeneratedVar that is used to figure out if a variable was generated internally (by the back end)
* [prepared] After doing casualization no new alias variable will be introduced (substitutions will be still made) by the back end. This will be needed once the system get split up right after the casualization module into systems for initialization and simulation.

### DIFF
--- a/Compiler/BackEnd/BackendDAEUtil.mo
+++ b/Compiler/BackEnd/BackendDAEUtil.mo
@@ -8269,21 +8269,21 @@ algorithm
   comp := BackendDAE.SINGLEEQUATION(eqIdx,varIdx);
 end makeSingleEquationComp;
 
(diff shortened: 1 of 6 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2015-06-05_14-18-45](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/77/console)

PR #39, opened 2015-06-14
Refactor enumeration in SCode module

gossen:
Patch replaces transformation Absyn.ENUMERATION --> SCode.ENUMERATION --> SCode.CLASS to Absyn.ENUMERATION --> SCode.CLASS and fix bug https://trac.openmodelica.org/OpenModelica/ticket/3327.
Also SCodeDumpTpl is improved to reduce trailing and double whitespaces in its output.


sjoelund:
See OpenModelica/OpenModelica-testsuite#28 for a comment on why detecting things at Absyn to SCode is often a bad thing.

PR #40, opened 2015-06-15
HPC-OM profiling changes

marchartung:
Hudson gives me a make-error ( No rule to make target `git-sanity-check'.) and just shows me 2 files changed.
So here the pull request.


sjoelund:
This is why you should always do a pull request and look at the changes you are actually proposing. Your changes propose to remove the git-sanity-check target from the Makefile.

sjoelund:
The changes would also break the new documentation by reverting to the old default_profiling.xsl

sjoelund:
You are also enforcing cmake 2.8.11 when we do not have it for all build targets...

sjoelund:
You also reverted the changes in Compiler/runtime/SimulationResults.c :(

PR #48, opened 2015-06-18
Introduce new API call listFile

sjoelund:
The function takes a class name and only prints elements that
originated in the same file as this class. It also adds a within
statement at the top of the file.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2015-06-18_15-54-57](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/162/console)

PR #63, opened 2015-06-22
Fix wrong parameter name for regex

sjoelund:
Was behaviour of "ignore case"=true, but said insensitive=true.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2015-06-22_16-12-35](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/193/console)

PR #77, opened 2015-06-26
Inverse solving of algorithms

lochel:
Solve for inputs (or a combination of inputs and outputs) of an algorithm section if necessary

Therefore, the algorithm is treated as a nonlinear loop. This should be improved for discrete or hybrid systems (mixed continuous and discrete systems).
See #3081


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2015-06-26_14-59-18](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/233/console)

PR #106, opened 2015-07-06
Add failtrace flag to SimCodeUtil.createSimCode function

gossen:
The error message contains source info for function createSimCode and appears in some tests,
so changes in SimCodeUtil.mo may cause changes in tests output.
This patch add failtrace flag to createSimCode error message to hide it.

### DIFF
--- a/Compiler/SimCode/SimCodeUtil.mo
+++ b/Compiler/SimCode/SimCodeUtil.mo
@@ -473,7 +473,9 @@ algorithm
     then (simCode, (highestSimEqIndex, equationSccMapping));
 
     else equation
-      Error.addInternalError("function createSimCode failed [Transformation from optimised DAE to simulation code structure failed]", sourceInfo());
+      if Flags.isSet(Flags.FAILTRACE) then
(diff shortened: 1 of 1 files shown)


sjoelund:
The sourceInfo() should not really be removed. Instead, the error should simple never appear or use the source information from the equation.

gossen:
This error message can be hidden using failtrace flag.

lochel:
I think the SimCode warning shouldn't be displayed if the initialization failed, since there should be always a proper error message from the initialization module.
But once I moved the initialization to the middle of the back end transformation, this will be obsolete anyway.

lochel:
To make this clear: I don’t think it is a good idea to put this error message on a flag. SimCode should never fail, and therefore the error message can be just there. The only odd thing here is, that the initialization is performed in SimCode.
It would be fine to hide this warning if initialization fails, but otherwise it should be just there.

gossen:
I don't think it can be done easily.

lochel:
Probably, xou could do something like lochel/OMCompiler@470defb50.

lochel:
It passed already the test suite together with lochel/OpenModelica-testsuite@6265190d.

PR #108, opened 2015-07-07
Documentation for C++ Simulation Runtime

MarcusWalther:
The documentation of the C++ Simulation Runtime is automatically generated using doxygen. For this purpose we need some custom css/html/image source-files, but it's not allowed to upload them. How should I add my commit to the repository? 
I know there is the doc-submodule, but I'm not sure if this kind of automatically generated documentation should be part of this module, because it is only generated if the user passes a command to cmake (see SimulationRuntime/cpp/CMakeLists.txt).


adrpo:
I don't know how we should handle this. 
As far as I know we have a Hudson job that builds the documentation and we could add the building of CPP documentation into that job and push it online into: https://openmodelica.org/doc/.
Unfortunately @sjoelund is in holidays but hopefully he'll have some time in the next days to give some opinion about this.

MarcusWalther:
@sjoelund can you please give an advice so that I can close the request? Thank you.

sjoelund:
Generated files should never be part of the repository. Why do you need to add custom css, html, and images? The OM logotype in particular is already part of the documentation repository and should of course not be duplicated. It's even available as svg, so the png in this commit should be avoided at all costs. The header, footer, and css look generic and could be put in the documentation package as well and shared between all subprojects using doxygen (to get a uniform look). In that case, makefiles should be part of the documentation repository as well (and not use cmake).

MarcusWalther:
Ok Martin, I will move everything to the doc-project. I will place the doxygen layout stuff into doc/doxygen, so everyone can use it and we can get the same look and feel. 
But I prefer to use cmake for the building process, because in the doxygen-configuration are some placeholder variables. To fill them, I either need cmake or autoconf and it's a mess to get these things running on windows and linux with autoconf, so cmake is the way to go.

PR #109, opened 2015-07-08
Synchronous improvements

gossen:
* Collapse all continuous and unspecified partitions into single continuous-time partition;
* Print base clocks in synchronous dumps;
* Do partitioning for removed equations. We need that for handling such constructs:
```modelica
when Clock() then
  Modelica.Utilities.Streams.print("clock ticks at = " + String(sample(time)));
end when;
```
The equation ordering and output format is changed, so many tests are affected

### DIFF
--- a/Compiler/BackEnd/BackendDAE.mo
+++ b/Compiler/BackEnd/BackendDAE.mo
@@ -70,13 +70,15 @@ type EqSystems = list<EqSystem>;
 public
 uniontype EqSystem "An independent system of equations (and their corresponding variables)"
   record EQSYSTEM
-    Variables orderedVars "ordered Variables, only states and alg. vars";
-    EquationArray orderedEqs "ordered Equations";
+    Variables orderedVars                   "ordered Variables, only states and alg. vars";
+    EquationArray orderedEqs                "ordered Equations";
     Option<IncidenceMatrix> m;
     Option<IncidenceMatrixT> mT;
     Matching matching;
-    StateSets stateSets "the state sets of the system";
+    StateSets stateSets                    "the state sets of the system";
     BaseClockPartitionKind partitionKind;
+    EquationArray removedEqs               "these are equations that cannot solve for a variable.
+                                            e.g. assertions, external function calls, algorithm sections without effect";
   end EQSYSTEM;
 end EqSystem;
 
@@ -115,7 +117,6 @@ uniontype Shared "Data shared for all equation-systems"
                                              data about variables' names, comments, units, etc. is preserved as well as
                                              pointer to their values (trajectories).";
     EquationArray initialEqs                "Initial equations";
-    EquationArray removedEqs                "these are equations that cannot solve for a variable. for example assertions, external function calls, algorithm sections without effect";
     list< .DAE.Constraint> constraints     "constraints (Optimica extension)";
     list< .DAE.ClassAttributes> classAttrs "class attributes (Optimica extension)";
     FCore.Cache cache;
--- a/Compiler/BackEnd/BackendDAECreate.mo
+++ b/Compiler/BackEnd/BackendDAECreate.mo
@@ -135,12 +135,11 @@ algorithm
   ieqnarr := BackendEquation.listEquation(ieqns);
   einfo := BackendDAE.EVENT_INFO(timeEvents, whenclauses_1, {}, {}, {}, 0);
   symjacs := {(NONE(), ({}, {}, ({}, {})), {}), (NONE(), ({}, {}, ({}, {})), {}), (NONE(), ({}, {}, ({}, {})), {}), (NONE(), ({}, {}, ({}, {})), {})};
-  outBackendDAE := BackendDAE.DAE(BackendDAEUtil.createEqSystem(vars_1, eqnarr)::{},
+  outBackendDAE := BackendDAE.DAE(BackendDAEUtil.createEqSystem(vars_1, eqnarr, {}, BackendDAE.UNKNOWN_PARTITION(), reqnarr)::{},
                                   BackendDAE.SHARED(knvars,
                                                     extVars,
                                                     aliasVars,
                                                     ieqnarr,
-                                                    reqnarr,
                                                     constrs,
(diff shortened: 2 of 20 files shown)


lochel:
What is the reason to collapse all continuous and unspecified partitions into single continuous-time partition?

gossen:
There are three reasons.

First, the specification says so. See 16.7.3:
> Other partitions where none of the variables in the partition are associated with any of the operators above have an
unspecified partition kind and are considered continuous-time partitions.
> All continuous-time partitions are collected together and form “the” continuous-time partition.

Second, I don't see any benefits from having multiple continuous and unspecified partitions.

The third and main reason is event handling.
Events, triggered in  one partition, should not propagate to other  partitions. But events in continuous or unspecified partition should propagate to other continuous or unspecified partition. This can be achieved by collapsing that partitions into single continuous partition.

sjoelund:
Multiple continuous-time partitions are useful for the simulation runtime:
* You can parallelise them trivially (they are independent)
* You can speed up the backend since the partitioned systems are smaller and the backend does not scale linearly

gossen:
You have a point. I'll remove that collapsing part.

casella:
I'm sorry, but I would definitely keep the collapsing part. As gossen noticed, the fact that the spec says so ("All continuous-time partitions are collected together and form “the” continuous-time partition.") is a good enough reason by itself. 
The second good reason was also mentioned by gossen, i.e., events in the continuous partition should be propagated to "the" whole partition - this is an automatic consequence of sticking to the spec. 
The third good reason is that Common Subexpression Elimination must span the entire set of continuous-time equations. If there are multiple partitions, the same expression might turn up multiple times in the generated code, which is just a waste of CPU time. In fact, having one partition is crucial for efficient handling of the IF97 water/steam model in Modelica.Media, which is designed to be efficient only if full CSE is performed. We already saw that if there are multiple continuous partitions, multiple redundant calls to Modelica.Media.Water.IF97_Utilities.waterBaseProp_ph() turn up in the code, which is very bad for performance of any thermo-fluid model using water, see the discussion in ticket #3102. Ad-hoc solutions were proposed in that ticket for this last issue, but IMHO, the right thing to do is just to stick to the spec and have only one continuous partition.

bernhard-thiele:
Maybe it is also possible with limited effort to keep both possibilities by introducing a flag. It would be interesting to investigate if significant performance can be gained for some real-world models if we manage to parallelize different partitions. Even if this is no good idea for thermo-fluid models, it might be good for others. The default behaviour could be to join all continuous-time partitions.

gossen:
I like the idea of @bernhard-thiele with two variants of partitioning.

casella:
IMHO, task parallelization should be performed by proper dependency analysis, task merging and clustering, etc. It shouldn't be a byproduct of something that was meant for partitioning clocked variables. Anyway, as long as this is not the default option, that's perfectly fine for me

lochel:
The partitioning is not just used for synchronous features. Originally, it was introduced to split up the system for TLM. Therefore it is needed to keep all the independent partitions. Or should it no longer be supported.
CSE could be implemented properly even if the system has multiple continuous-time partitions. That is no big deal and should not be the reason for this decision.
In general I don’t like to have everything on a flag, since this makes it really complicated to use omc efficiently.

sjoelund:
The partitioning was introduced before synchronous features were considered. And the same expression cannot occur in multiple partitions because then the subsystems are not independent. I suppose the only way they could occur in multiple subsystems is if they have parameter or constant variability (after optimizations), in which case they should be evaluated only once anyway.

casella:
To @sjoelund: consider this simple example

model foo
...
equation
  der(x) = ...;
  z = f_very_time_consuming(x);
  w = f_very_time_consuming(x);
  s = f_very_time_consuming(x);
end foo;

As states are known at each time step, the three algebraic equations will end  up in three different partitions, and the very time consuming function computed three times, which in general is not a good idea. So, it's not only parameter and constant variability variables, but also states.

Please note Modelica.Media produces this kind of situations all the time.

So, I'm fine with partitions, but then CSE should be applied at the whole system level, not at the partition level. This is a fundamental requirement for efficiency.

lochel:
I propose that the cse modul will collapse these three partitions. Right now I am not sure what is already available. If this is not in the master branch yet, it should be available soon.

lochel:
BTW: The example from above will currently just end up in a single partition. 
```modelica
pre-optimization module clockPartitioning:


unspecified partition
========================================

Variables (4)
========================================
1: y3:VARIABLE() .TestCSE.Test4, .Real type: Real []
2: y2:VARIABLE() .TestCSE.Test4, .Real type: Real []
3: y1:VARIABLE() .TestCSE.Test4, .Real type: Real []
4: x:VARIABLE() .TestCSE.Test4, .Real type: Real []


Equations (4, 4)
========================================
1/1 (1): der(x) = 1.0 / (1.0 + time)    [dynamic]
2/2 (1): y1 = TestCSE.f_very_time_consuming(x)    [dynamic]
3/3 (1): y2 = TestCSE.f_very_time_consuming(x)    [dynamic]
4/4 (1): y3 = TestCSE.f_very_time_consuming(x)    [dynamic]
```

Is that intended?

bernhard-thiele:
@lochel and @casella : I would have expected that if a `x` of a `der(x)` appears in another equation, that the corresponding equations are collected in the same partitions.

That seems particularly a necessary condition if considering "Clocked Discretized Continuous-Time Partition".

lochel:
Ahh, sorry for that. You are right (see 16.7.3 in Modelica spec.). That’s how I implemented the base-clock partitioning ;-)

gossen:
I 'll move removed equations partitioning to another pull request.

The collapsing of continuous partition is another issue, and may be considered later.

casella:
I had one case of duplicate CSE expressions due to multiple partitions which is described here: https://trac.openmodelica.org/OpenModelica/ticket/3102#comment:19
I am not sure what the root cause was, but I talked to Lennart, he'll investigate it and see whether collapsing different partitions is relevant or not in this case.

PR #111, opened 2015-07-09
changes to support Mingw64

adrpo:


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2015-07-10_01-50-35](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/360/console).

sjoelund:
Note that printf macros should cast to long or int explicitly since they expect the input to be of a certain length. %lu expects a long.

adrpo:
Or we could have different formatting strings depending on 32/64 bit. I'll deal with this.

sjoelund:
Should not use different strings for this. Just cast to long and use long printf

adrpo:
Why, if long is not always sizeof(void*). On Windows 64bit long long is sizeof(void*) so it wouldn't work properly.

sjoelund:
Use %p if you want to print a pointer? Or the long long directive always.

PR #114, opened 2015-07-14
Fix appearance of config flags for sphinx documentation

lochel:
@casella noted that the config flags are not displayed correctly for the [sphinx documentation](https://www.openmodelica.org/doc/OpenModelicaUsersGuide/latest/omchelptext.html?highlight=removesimpleequations#omcflag-removesimpleequations), since "--" is converted to a long dash.
@adrpo, can you please verify my fix, since I cannot build the documentation at the moment.

### DIFF
--- a/Compiler/Util/Flags.mo
+++ b/Compiler/Util/Flags.mo
@@ -2314,15 +2314,19 @@ protected function printConfigFlagName
 algorithm
   (outString,longName) := match(inFlag)
     local
-      String name, shortname;
+      String name, shortname, str;
 
     case CONFIG_FLAG(name = name, shortname = SOME(shortname))
       equation
-        shortname = if sphinx then "-" + shortname else Util.stringPadLeft("-" + shortname, 4, " ");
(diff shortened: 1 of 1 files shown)


adeas31:
This is not working. In fact it also breaks the cross reference. The problem is that sphinx is using smartypants which automatically converts the double dash to em-dash. And currently there is no way to escape the smartypants in sphinx (https://github.com/sphinx-doc/sphinx/issues/498).

Only solution is to disable the smartypants in `conf.py` by setting `html_use_smartypants` to `False`.

sjoelund:
I guess we can turn off the smarty pants. It's easy to use a Unicode em or en dash when needed anyway.

adeas31:
Easiest is to turn off SmartyPants OR if someone is really in favor of keeping it then we can wrap the flags in parsed-literal blocks because Sphinx do not apply SmartyPants in parsed-literal blocks.

lochel:
Okay, then we should just switch off SmartyPants.

PR #117, opened 2015-07-16
Stop pre/post-optimization on failure properly

lochel:
Currently, the pre/post-optimization phase will be run completely, even if a module failed already. That leads to strange and confusing error messages, which are not relevant, since a module failed already before.

### DIFF
--- a/Compiler/BackEnd/BackendDAEUtil.mo
+++ b/Compiler/BackEnd/BackendDAEUtil.mo
@@ -6594,7 +6594,7 @@ algorithm
   end if;
 
   // pre-optimization phase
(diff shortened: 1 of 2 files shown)


lochel:
BTW: I also changed the ``+d=optdaedump`` a bit. It will now show the matching information after each post-optimization module. That was not the case yet, which was a bit weird.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2015-07-16_09-53-26](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/381/console).

PR #120, opened 2015-07-17
Revised singularity check for linear systems

lochel:
* add explanation of dgesv return values
* let transformation fail again if "LINEAR_SYSTEM_INVALID" error message is triggered

### DIFF
--- a/Compiler/BackEnd/BackendDAETransform.mo
+++ b/Compiler/BackEnd/BackendDAETransform.mo
@@ -299,8 +299,11 @@ algorithm
       (jac, shared) = SymbolicJacobian.calculateJacobian(vars_1, eqns_1, m, true, ishared);
       // Jacobian of a Linear System is always linear
       (jac_tp, jacConstant) = SymbolicJacobian.analyzeJacobian(vars_1, eqns_1, jac);
-      // if constant check for singular jacobian
-      true = analyzeConstantJacobian(jacConstant, jac, arrayLength(mt), var_lst, eqn_lst, shared);
+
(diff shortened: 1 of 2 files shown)


lochel:
(@vwaurich, @wibraun) Or should the transformation continue anyway? If so, then this error message (LINEAR_SYSTEM_INVALID) should also become a warning.

vwaurich:
Possibly it should be a warning that the ith value is illegal whereat the ith value should also be displayed. I have no idea in which case an invalid jacobian entry occures at that compilation stage but I see no reason to abort translation although translation could continue.

lochel:
Okay. I will change it that way. I will also try to setup a model that runs into that case, since there is no such test in the test suite. 

----- Ursprüngliche Nachricht -----
Von: "vwaurich" <notifications@github.com>
Gesendet: ‎17.‎07.‎2015 16:01
An: "OpenModelica/OMCompiler" <OMCompiler@noreply.github.com>
Cc: "Lennart Ochel" <lennart.ochel@fh-bielefeld.de>
Betreff: Re: [OMCompiler] Revised singularity check for linear systems (#120)

Possibly it should be a warning that the ith value is illegal whereat the ith value should also be displayed. I have no idea in which case an invalid jacobian entry occures at that compilation stage but I see no reason to abort translation although translation could continue.

lochel:
@vwaurich Well, I also do not see how a system should fail due to invalid input. Hence, I will keep it an error message and let it abort the translation. I guess that case is just to cover all the valid return values of dgesv, even if they will not occur in our environment anyway.

PR #128, opened 2015-07-27
Changed number of event indicators for c++ based FMU export

MarcusWalther:
Because the c++-runtime handles events different than the c-runtime, the number of event indicators created by FMUCommon.tpl was not correct for both runtimes. Therefore i've added a switch inside the new function "getNumberOfEventIndicators()".

This fixes a segfault, that occurs when the CoupledClutches model is translated to FMU and simulated in OpenModelica. 

@rfranke  Can you please take a look at the request and test if these changes will also fix your local segfault?

### DIFF
--- a/Compiler/Template/CodegenCppInit.tpl
+++ b/Compiler/Template/CodegenCppInit.tpl
@@ -63,7 +63,7 @@ template fmiModelDescriptionAttributes(SimCode simCode, String guid)
       let generationDateAndTime = CodegenFMUCommon.xsdateTime(getCurrentDateTime())
       let variableNamingConvention = 'structured'
       let numberOfContinuousStates = vi.numStateVars
-      let numberOfEventIndicators = eventIndicatorsLength(simCode)
+      let numberOfEventIndicators = CodegenFMUCommon.getNumberOfEventIndicators(simCode)
       <<
       fmiVersion="<%fmiVersion%>"
       modelName="<%modelName%>"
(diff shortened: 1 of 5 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2015-07-27_15-34-56](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/436/console).

rfranke:
The segfaults started with the changes to Kinsol -- somewhen around https://github.com/OpenModelica/OMCompiler/commit/2542dd936a6b26b7e227978fb723b2b72dffbc5f

They disappear with https://github.com/OpenModelica/OMCompiler/commit/d9b26e6b6fb2b6efa951713889794202088303cd as the Newton solver is used per default now.

MarcusWalther:
Thank you for the feedback, I have not expected that d9b26e6b6fb2b6efa951713889794202088303cd fixes something. However, I will upload the files and change the CoupledClutches FMU 2.0 test to verified simulation.

rfranke:
Why does the Cpp runtime handle events differently? 

Couldn't the event handling in the Cpp runtime be unified with the event handling of the C runtime, in order to avoid cluttering of the code generation?

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2015-07-28_07-42-17](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/438/console).

MarcusWalther:
I think this can be explained in detail by @wibraun and @niklwors . I just know that a switch has been added in the sim code generation, that handles the event things different for c and c++ code target. The c-runtime generates complex event indicators like "((a || b) & (c > d))". The c++ code generation is only able to generate simple event indicators, like "(a || b)" or "(c > d)". Thus the complex expressions must be splitted for the c++ code generation. I'm not sure how easy it is to change the behaviour of the c++ code generation.

niklwors:
We will also generate the zero crossing functions like the c- runtime in the near future.

PR #157, opened 2015-08-11
Move threadData away from simulation data

sjoelund:
Each thread needs its own threadData, and must be initalized. By moving
the data, it will be easier to see if the correct threadData is passed.
This resolves issues with HPCOM using the same thread-local data for all
threads.

Move the generated function prototypes to a common header since having
the function prototypes to a common header gives compiler errors if
there is a mismatch, instead of run-time errors.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2015-08-11_18-59-02](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/575/console).

PR #158, opened 2015-08-12
Fixes for listFile

sjoelund:
listFile now lists non-class elements with a different source file
than the given class. The reason is that components always need to be
part of the same file.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2015-08-12_21-36-29](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/588/console).

PR #175, opened 2015-08-26
Fix for return in try-block

sjoelund:
This fixes bug #3240, where the jmp_buf is set to a deeper stack
frame after returning from the middle of a try block (the jmp_buf
needs to be restored after exiting the block).


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2015-08-26_09-33-40](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/646/console).

PR #190, opened 2015-09-08
Fix for #3409.

perost:
- Type cast matrix elements before assembling the matrix,
  not after when it's too late.
- Various type fixes for elaboration of matrices.
- Fix List.mapFoldList so that it updates the argument,
  not just reusing the input argument for each call.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2015-09-08_19-06-38](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/728/console).

PR #194, opened 2015-09-09
Fix for #3421.

perost:
- When checking if the rhs is a function call in an tuple assignment,
  check the unelaborated expression since the elaborated one might no
  longer be a function call.
- Changed Types.stripTypeVars to also strip tuple element names
  (for less confusing error messages).


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2015-09-09_12-59-48](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/732/console).

PR #196, opened 2015-09-10
When merging annotation, consider the order

sjoelund:
Previously, annotations did not consider the order of submods.
These are now considered such that code does not move around when
adding graphical annotations. If you want to change the order
explicitly, this can be done by first removing the old annotation.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2015-09-10_13-26-58](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/750/console).

PR #232, opened 2015-10-13
rewritten solveSimpleEquation

vruge:


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2015-10-13_18-40-22](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/928/).

lochel:
Is this just about cleaning up the implementation or does it also change the behaviour of the module?

vruge:
It's not just cleaning up of the implementation. But I'm not sure that not just cleaning prefaces change of the behaviour of the module

lochel:
Please add a description to the module (e.g. a detailed comment above the entry point), since it is not clear what is actually treated as "simple equation" and so on.

vruge:
possible. At the beginning I I will rename the module in solveSingleEquation.

PR #234, opened 2015-10-14
 - use functionTree from initDAE further for simDAE

wibraun:


lochel:
I thought about merging the function trees instead of just passing it through to keep initialization and simulation independent, which would make it possible to run those tasks in parallel.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2015-10-14_13-14-06](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/941/console).

PR #242, opened 2015-10-19
Fix inst of Annex60 (and maybe Buildings).

perost:
- Type convert record values when when evaluating modifier values,
  to make sure they are of the correct types.
- Better check for duplicate modifiers,
  gets rid of some "duplicate modifier" errors.
- Better merging of record modifiers,
  gets rid of even more "duplicate modifier" errors.
- Some cleanup.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2015-10-19_17-34-27](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/976/console).

PR #244, opened 2015-10-20
Use va_copy when re-using a va_list

sjoelund:
This fixes a bug in GC_asprintf, which is now used more often than
earlier. This should fix simulation of models on OSX.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2015-10-20_16-56-03](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/991/console).

PR #258, opened 2015-10-28
Fixed alias default and initial equation sub expression tag

paul-scott:
Fixes two differences between exported XML files and the format standard (what JModelica and casadi produce/expect).

The first is that when there are no alias' a blank string was used instead of "noAlias" :
https://svn.jmodelica.org/trunk/XML/jmodelicaScalarVariable.xsd

The second was that the InitialEquations section was using the tag of <equ:Sub> instead of <exp:Sub>.

After making these changes the casadi xml parser can now load an exported OM XML file without throwing an exception.

I'm not sure if any of the unit tests need to be modified because of this change.  This is my first pull request for this project so I'm not sure how to initiate the Hudson testing on the branch.

### DIFF
--- a/Compiler/Template/CodegenXML.tpl
+++ b/Compiler/Template/CodegenXML.tpl
@@ -218,7 +218,7 @@ template getAliasVarXml(AliasVariable aliasvar)
  "Returns the alias Attribute of ScalarVariable."
 ::=
 match aliasvar
-  case NOALIAS(__) then ""
+  case NOALIAS(__) then "noAlias"
   case ALIAS(__) then '<%crefStrXml(varName)%>'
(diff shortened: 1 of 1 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2015-10-28_09-16-03](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1056/console).

adrpo:
Hi @paul-scott. Only certain users can run the testsuite on the pull requests. @sjoelund has now ran the testsuite on your pull request and everything was fine, so this pull request is now merged.
Thanks for your contribution!

PR #267, opened 2015-11-03
Use equivTypes function instead of equality

sjoelund:
This fixes #3518. Equality does not take into consideration that the
source information might be different for two types.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2015-11-03_13-40-51](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1101/console).

PR #270, opened 2015-11-04
Fix for Buildings.

perost:
- Don't reorder elements in Inst.instElementList, to avoid record
  elements being reordered based on their modifications. Unless
  the elements belong to a function, then the elements are sorted
  as before to avoid things breaking.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2015-11-04_14-00-02](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1109/console).

PR #272, opened 2015-11-05
Merge handling of pre/post-optimization modules to make the back end more flexible

lochel:
Optimization modules are no longer allowed to fail. If it is intended that a module should fail for some reason, then that need to be handled within the module from now on.
New modules are now defined using a function pointer to the module implementation, a name as string, and a Boolean that defines if that module is "post-optimization only" or not.

### DIFF
--- a/Compiler/BackEnd/BackendDAEFunc.mo
+++ b/Compiler/BackEnd/BackendDAEFunc.mo
@@ -44,17 +44,11 @@ public import BackendDAE;
 /*************************************/
 
 public
-partial function preOptimizationDAEModule "
-  This is the interface for pre-optimization modules."
+partial function optimizationModule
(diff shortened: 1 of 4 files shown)


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2015-11-05_15-38-32](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1116/).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2015-11-05_16-23-46](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1117/console).

PR #277, opened 2015-11-08
+ Fix handling of complex equations so that we won't generate unecess…

mahge:
+ Fix handling of complex equations so that we won't generate unecessary nonlinear systems.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2015-11-08_15-11-03](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1123/console).

PR #289, opened 2015-11-14
Fix for #2959.

perost:
- Propagate partial attribute to environment scope elements.
- Add SourceInfo to Prefix.ComponentPrefix.
- Replace Boolean message argument to Lookup.lookupClass with
  Option<SourceInfo> so we know where an error comes from.
- Print error in Lookup.lookupClassQualified2 if we're trying
  to look a name up in a partial class.
- Propagate redeclares when instantiating an element of an empty
  array in InstVar.instArray, so we get the correct type.

### DIFF
--- a/Compiler/FFrontEnd/FGraph.mo
+++ b/Compiler/FFrontEnd/FGraph.mo
@@ -122,6 +122,22 @@ algorithm
   outRef := listHead(currentScope(inGraph));
 end lastScopeRef;
 
+public function setLastScopeRef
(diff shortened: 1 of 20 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2015-11-14_14-41-18](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1173/console).

PR #300, opened 2015-11-19
adding context depending extrapolation

wibraun:
That breaks currently for some unknown reason 2 example of ThermoSysPower, 
but it fixes partially #3111 and fully #3544.

### DIFF
--- a/SimulationRuntime/c/simulation/solver/dassl.c
+++ b/SimulationRuntime/c/simulation/solver/dassl.c
@@ -128,34 +128,6 @@ dummy_precondition(int *neq, double *t, double *y, double *yprime, double *savr,
 
 static int continue_DASSL(int* idid, double* tolarence);
 
-enum EVAL_CONTEXT
-{
-  CONTEXT_UNKNOWN = 0,
-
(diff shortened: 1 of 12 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2015-11-19_23-03-29](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1216/console).

casella:
I've seen some cases where extrapolation can be dangeorus. For example, you have a pressure source, a valve, a pipe, and a pressure sink. The valve is closed with a ramp opening signal. Until the valve is almost closed, the loss of the pipe dominates, so the valve downstream pressure (which is the unknown of a nonlinear system) stays close to the source pressure. Only when the valve opening approaches zero, the valve downstream pressure suddenly drops. If the time steps are not small enough, extrapolation can actually bring the estimated pressure value to negative values, from which the Newton solver can hardly recover. Maybe the problem with the ThermoSysPro is of this nature.

PR #313, opened 2015-11-30
Performance changes

sjoelund:


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2015-11-30_14-47-53](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1277/).

vruge:
differentiateEqnsLst:
I would try to move the case "// Why do we just return an empty set if we somehow fail?" in the for loop

sjoelund:
But then the code does not do the same thing it did before. I think the breaking changes are due to trying to change the default for simplify.

vruge:
You're dead right!

PR #328, opened 2015-12-09
Fix scaling issues for initialization

sjoelund:
Iterate over the smaller set of variables.

### DIFF
--- a/Compiler/BackEnd/BackendVariable.mo
+++ b/Compiler/BackEnd/BackendVariable.mo
@@ -2939,6 +2939,42 @@ algorithm
   end try;
 end traversingVarIndexFinder;
 
+public function getVarIndexFromVariablesIndexInFirstSet
+  input BackendDAE.Variables inVariables;
+  input BackendDAE.Variables inVariables2;
+  output list<Integer> v_lst;
+protected
+  array<list<Integer>> a;
+algorithm
+  (a,_) := traverseBackendDAEVars(inVariables,
+    function traversingVarIndexInFirstSetFinder(inVars = inVariables2), (arrayCreate(1,{}),arrayCreate(1,1)));
+  v_lst := listReverse(a[1]);
+end getVarIndexFromVariablesIndexInFirstSet;
+
+protected function traversingVarIndexInFirstSetFinder
+"author: Frenkel TUD 2010-11"
+  input BackendDAE.Var inVar;
+  input BackendDAE.Variables inVars;
+  input tuple<array<list<Integer>>,array<Integer>> inIndices;
+  output BackendDAE.Var outVar = inVar;
+  output tuple<array<list<Integer>>,array<Integer>> outIndices;
+protected
+  DAE.ComponentRef cr;
+  list<Integer> indices1,indices2;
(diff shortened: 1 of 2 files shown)


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2015-12-09_15-59-50](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1324/).

lochel:
That's not working, since the element indices based on `inEqSystem.orderedVars` is relevant.

sjoelund:
Then it should be sorted in some other way afterwards, because this is causing an O(N^2) scaling for an algorithm that should be O(N) -- it's really slow in the larger models.

lochel:
That should be possible, but not using `BackendVariable.getVarIndexFromVariables`.

lochel:
I can do that with another pull request. Or are you going to fix it?

sjoelund:
I might fix it tomorrow if I have time to spare.

sjoelund:
I am now running the tests using the above approach.

lochel:
Good - that should work.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2015-12-10_10-34-35](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1339/console).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2015-12-10_10-48-17](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1340/console).

lochel:
How big is the impact of this improvement?

sjoelund:
I will know in about an hour :)

sjoelund:
Before: Performance of analyzeInitialSystem (initialization): time 1831
Now: Performance of analyzeInitialSystem (initialization): time 485

sjoelund:
Of the remaining time, most of it is not relating to this fix, but rather the sorting+matching/etc of the largest system of equations in the initialization problem:

Performance of fixInitialSystem (initialization 2) [nEqns: 49146, nAddEqs: 4608, nAddVars: 0]: time 433.4
(Which is just the last part of one of the >20000 iterations)

So I am now sure this part of the code now scales O(N) instead of O(N^2).

lochel:
Well, the example seems to be under-determined. Therefore, it will not scale {{{O(N)}}}. First, I think, we should focus on well-determined models.

casella:
This has been solved thanks to Volker's analysis. Some parameters eventually multiplying derivatives were accidentally set to zero, and evaluateParameters removed some derivatives, so that there were a few more initial equations than states. After correcting this, the initial system became square and the time for analyzeInitialSystem dropped from 335 to 40 seconds, without using the latest commits of this branch, due to problems with nightly builds. We expect further improvements with the latest nightly, thanks to Martin's contribution.

casella:
Martin's improvements slashed the execution time of analyzeInitialSystem by 50% to 20 seconds. Excellent job!

PR #347, opened 2016-01-08
Add a simple Modelica diff based on tree parsing

sjoelund:
There is a hand-written parse tree generator for a relaxed version of
the Modelica grammar.

There exists an algorithm for tree merging, which is based on trying
to diff all the subtrees in a node, to figure out which subtrees
changed (syntactically, not lexically). Only these subtrees are then
traversed.

Diff algorithms scale O(N*D), which made the old scanner-based diff
slow since N and D are both very large (many whitespace changes).
The new diff has a low N (because entires subtrees are pruned away
early on), and a low D (because whitespace changes are often pruned
away in semantically identical subtreees).

However, the algorithm still scales poorly if there is a large number
of nested classes in the file. It works much better when models are
stored in separate files (takes around 1-2 seconds for 300 kB files).
The lexer-based diff did spend hours to perform this diff however...

The diff is patched to perform move operations in order to preserve
whitespace better (otherwise whitespace would be reset to the updated
file).

Token content is performed using a tokenContentEq operation, which
can do string comparison on substrings directly (to avoid creating
strings only to compare them). The (Boolean) tree comparison is
linear in both memory and time.

There is a labelling of some nodes (like named modifications), in order
to know which subtrees belong together in a comparison.

Whitespace and indentation is considered in a similar way as the old
comparison algorithm.

### DIFF
--- a/Compiler/Lexers/LexerModelicaDiff.mo
+++ b/Compiler/Lexers/LexerModelicaDiff.mo
@@ -33,22 +33,27 @@ end LexTable;
 function scan "Scan starts the lexical analysis, load the tables and consume the program to output the tokens"
   input String fileName "input source code file";
   output list<Token> tokens "return list of tokens";
+  output list<Token> errorTokens;
 protected
   String contents;
 algorithm
   contents := System.readFile(fileName);
-  tokens := lex(fileName,contents);
+  (tokens, errorTokens) := lex(fileName,contents);
 end scan;
 
 function scanString "Scan starts the lexical analysis, load the tables and consume the program to output the tokens"
   input String fileSource "input source code file";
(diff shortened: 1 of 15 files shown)


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2016-01-08_13-50-39](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1412/).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-01-08_14-04-38](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1413/console).

PR #352, opened 2016-01-21
Fix for #3633.

perost:
- Don't generate result assignment when match is used as a
  non-returning call.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-01-21_14-04-15](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1445/console).

PR #368, opened 2016-02-01
Remove some matchcontinue

hkiel:


SimulationRuntime/c/util/modelica_string.c
```diff
@@ -199,6 +199,15 @@ modelica_string modelica_real_to_modelica_string_format(modelica_real r,modelica
...
 }
 
+/* Convert a modelica_string to a modelica_string, used in String(s) */
+
+modelica_string modelica_string_to_modelica_string(modelica_string s)
+{
+  void *res = alloc_modelica_string(MMC_STRLEN(s));
+  strcpy(MMC_STRINGDATA(res), MMC_STRINGDATA(s));
+  return res;
+}
```
sjoelund: Why not just return s? It's garbage collected, so this should not be a problem (or changing String(s) to just be a no-op?)

hkiel:
What if String("foo") is called? Then "foo" may be only temporary and the assignment may become invalid.

sjoelund:
It will then be in constant memory. There are no stack-allocated strings in OM

hkiel:
I updated the code. Can this be merged now? Has Hudson already checked this???

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-02-02_16-11-10](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1496/console).

PR #371, opened 2016-02-02
Add support for the trapezoidal method (RK2)

sjoelund:
Error for der(x)=1-x shrinks by halving the step size:
* Euler: x2
* Trapezoidal: x4
* RK4: x16

### DIFF
--- a/SimulationRuntime/c/simulation/solver/solver_main.c
+++ b/SimulationRuntime/c/simulation/solver/solver_main.c
@@ -67,14 +67,12 @@
 
 double** work_states;
 
-const int rungekutta_s = 4;
-const double rungekutta_b[4] = { 1.0 / 6.0, 1.0 / 3.0, 1.0 / 3.0, 1.0 / 6.0 };
-const double rungekutta_c[4] = { 0.0, 0.5, 0.5, 1.0 };
-
 typedef struct RK4_DATA
 {
   double** work_states;
   int work_states_ndims;
+  double *b;
+  double *c;
 }RK4_DATA;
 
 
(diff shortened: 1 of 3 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-02-02_12-21-58](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1493/console).

vruge:
Look like Heun's method, which has explicit and not implicit like trapezoidal rule. 
trapezoidal rule: is A-stable.
heun is not A-stable.
By the way lobatto2 is trapezoidal rule (maybe unlucky naming)

sjoelund:
Then fix the naming :)

casella:
We have two problems: one is using the right names, the other is aliases.
I agree with Vitalij we shouldn't call this trapezoidal, but rather heun
We have aliases: lobatto2 is trapezoid, radau1 is implicit euler, how to make this clear and not confusing for users?

sjoelund:
It's explicit trapezoidal :) But yes, @adeas31 OMEdit needs to once again show the tooltip for the description of the solver method as well (where it is easy to describe the aliases of the solvers)...

vruge:
How about an other mapping in OMEdit:
implict euler/(radua1) -> solve=radau1

or renmaing radau1 and labotto2...

adeas31:
Nothing is hard coded in OMEdit. Whatever we have in **simulation_options.h** is shown in OMEdit.

Properly populate the following arrays, 
**extern const char *SOLVER_METHOD_NAME[S_MAX];
extern const char *SOLVER_METHOD_DESC[S_MAX];**

and the solver name and description will be shown automatically in OMEdit.

sjoelund:
@adeas31 SOLVER_METHOD_DESC is no longer used in OMEdit. It is missing those nice descriptions (at least for me on Linux)

sjoelund:
@vruge perhaps it would be better to make the default names be impeuler and trapezoidal, with a description that it is lobatto/radau. I fixed the renaming of exp trapezoidal = Heun and added some descriptions to make things better. I should also fix the documentation generation as it is broken...

PR #393, opened 2016-02-10
attempt to make OMEdit work under Windows with Cpp

rfranke:
In OMEdit under Windows:
  - select Tools->Options->Simulation->Target Language: Cpp
  - select Tools->Options->Simulation->Target Compiler: gcc
  - open a model, e.g. Modelica.Blocks.Examples.Filter
  - simulate ->

This results in:

    Process crashed
    Simulation process failed. Exited with code -1073741515.

This is most likely because the path of the Cpp dll's is not set. It is:

$(omHome)\\lib\\omc\\cpp

Should this path possibly be added to OMCompiler/Compiler/Main/Main.mo?

### DIFF
--- a/Compiler/Main/Main.mo
+++ b/Compiler/Main/Main.mo
@@ -762,7 +762,7 @@ algorithm
         // do we have the correct libexec stuff?
         true = System.directoryExists(omHome + "\\mingw\\libexec\\gcc\\mingw32\\4.4.0");
         newPath = stringAppendList({omHome,"\\bin;",
-                                    omHome,"\\lib;",
+                                    omHome,"\\lib\\omc\\cpp;",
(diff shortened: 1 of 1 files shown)


adrpo:
@adeas31: didn't we added this to the path in OMEdit?

adrpo:
sorry, i closed it by mistake

adrpo:
I think we adding it to Main.mo will not help as OMEdit does things differently:
1. runs translateModel
2. runs make to build the model
3. runs the executable from OMEdit

rfranke:
You are right. This pull request does not explain why msvc paths are currently found. These find in StringHandler::simulationEnvironment() in

    OMEdit/OMEdit/OMEditGUI/Util/StringHandler.cpp

This function might be the right place to fix.

PR #402, opened 2016-02-12
Changed behaviour of getComponents API.

perost:
- Try to return the path where a type should be found, even if it
  doesn't actually exist.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-02-12_15-16-04](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1563/console).

PR #411, opened 2016-02-14
fix stack overflow

hkiel:
make functions Ceval.ceval and Types.matchTypeList use loops instead of recursion


sjoelund:
Why a try-else-fail? That's the same as not using a try-block.

hkiel:
try-else-fail was a relict from an earlier version.
Removed it (with --amend, but needed to do a merge because of more recent changes)

sjoelund:
Closing with #414 (git rebase -i origin/master to remove the duplicate commits with the same message, and the odd merge)

PR #14, opened 2015-06-03
added assert for pow(x,0.5) = sqrt(x)

vruge:

PR #13, opened 2015-06-03
- Fixed typo.

perost:

PR #12, opened 2015-06-01
IndexReduction

vruge:
- usig static state selction(dummyDerivatives) in dynOpt
- update heuristic
- cosmetic change
  - using for loop construct

PR #11, opened 2015-05-29
Adds serialization functions for MetaModelica values

modlfo:
Adds package Serializer that contains functions to serialize values to files.

PR #420, opened 2016-02-17
Don't allocate from GC while we are in a runtime loop.  It causes pro…

MichaelKobierski-OPALRT:
…blems if client runs the function from multithreaded environment (garbage collection fails).

### DIFF
--- a/SimulationRuntime/c/util/base_array.c
+++ b/SimulationRuntime/c/util/base_array.c
@@ -307,14 +307,24 @@ size_t calc_base_index(int ndims, const _index_t *idx_vec, const base_array_t *a
 
 size_t calc_base_index_dims_subs(int ndims,...)
 {
-
     int i;
     size_t index;
-
-    _index_t *dims = (_index_t*)omc_alloc_interface.malloc(sizeof(_index_t)*ndims);
-    _index_t *subs = (_index_t*)omc_alloc_interface.malloc(sizeof(_index_t)*ndims);
-
     va_list ap;
+
+    // Do not allocate from GC (preferably not at all) during runtime.
+    // GC has issues if client calls from transient threads.
+    enum { max_dims = 256 };
+    _index_t dims[max_dims];
+    _index_t subs[max_dims];
+    if((size_t)ndims > max_dims) {
+        FILE_INFO info = omc_dummyFileInfo;
+        omc_assert(NULL, info, 
+            "Dimension given (%d) is beyond built-in limit (%s)", 
+            ndims, max_dims);
+        return -1;
+    }
+    
+    
     va_start(ap,ndims);
     for(i = 0; i < ndims; ++i) {
         dims[i] = va_arg(ap, _index_t);


adrpo:
I don't think we can push this in because the array sizes are really not fixed.
Is also strange that this helps as we have plenty of other allocation by the GC in all the array functions.
Do you have a model that exposes this issue? What OS are you running on?

MichaelKobierski-OPALRT:
Hi Adrian, In practice, do you see your array dimensions going over 256?  I haven't observed it for any of my models.

I know the other array functions also allocate memory, but I didn't find any calls to them which result in a problem - only this one.  During simulation, GC calls abort, here.

win32_threads.c, line 1544
GC_INNER void GC_push_all_stacks(void) {
...
  if (!found_me && !GC_in_thread_creation)
    ABORT("Collecting from unknown thread"); // this is called
}

And a stack trace:
(gdb) where
#0  0x74e0338f in KERNELBASE!DebugBreak () from C:\Windows\syswow64\KernelBase.dll
#1  0x05e20d7a in GC_push_all_stacks () at win32_threads.c:1544
#2  0x05e20036 in GC_mark_some_inner (cold_gc_frame=0x468fa74 "túh\004,úh\004!Yá\005túh\004") at mark.c:416
#3  0x05e2022d in GC_mark_some (cold_gc_frame=0x468fa74 "túh\004,úh\004!Yá\005túh\004") at mark.c:525
#4  0x05e1dd21 in GC_stopped_mark (stop_func=0x5e1d10c <GC_never_stop_func>) at alloc.c:642
#5  0x05e1df3a in GC_try_to_collect_inner (stop_func=0x5e1d10c <GC_never_stop_func>) at alloc.c:461
#6  0x05e1e3d3 in GC_collect_or_expand (needed_blocks=1, ignore_off_page=0, retry=0) at alloc.c:1270
#7  0x05e1e597 in GC_allocobj (gran=1, kind=1) at alloc.c:1360
#8  0x05e1a421 in GC_generic_malloc_inner (lb=4, k=1) at malloc.c:132
#9  0x05e1a4dc in GC_generic_malloc (lb=4, k=1) at malloc.c:181
#10 0x05e1a865 in GC_core_malloc (lb=4) at malloc.c:284
#11 0x05db377c in calc_base_index_dims_subs () from C:\TheFMU_test\FMU\TheFMU.win32\binaries\win32\TheFMU.dll
#12 0x05d67610 in TheFMU_eqFunction_578 (data=0x8880a80, threadData=0x8b4a860) at TheFMU.c:1427
#13 0x05d6c25f in TheFMU_functionDAE (data=0x8880a80, threadData=0x8b4a860) at TheFMU.c:2538
#14 0x05d43a26 in TheFMU_fmiEventUpdate (c=0x69d5f70, intermediateResults=1 '\001',
    eventInfo=0x468fe00) at C:/OpenModelica1.9.3//include/omc/c/fmi1/fmu1_model_interface.c:723

And GC documentation on threads: 
For a multi-threaded program, some more rules apply:

    Files that either allocate through the GC or make thread-related calls should first define the macro GC_THREADS, and then include "gc.h". On some platforms this will redefine some threads primitives, e.g. to let the collector keep track of thread creation. 

In the application, fmiEventUpdate is called in an OpenMP parallel region, where threads are created and joined, invisible to GC.

In addition, you will have a performance improvement by using static arrays in place of dynamic. If you really need it to be dynamically allocated, let's put a check on ndims, if it is bigger than a reasonable number, allocate the array dynamically using a standard malloc (not GC), and free it at the end of the function.

How does that sound?

casella:
Q: In practice, do you see your array dimensions going over 256?
A: Absolutely! There are very interesting emerging applications where we could easily have thousands or possibly tens of thousands of elements in an array. Not millions, mind you, but definitely more than 256. Please make sure that the compiler scales well and do not make any a-priori assumptions on the size of models!

adrpo:
Yes, using normal malloc/free might work. 
You could also try to build the FMU using the CPP target as that doesn't use the Boehm GC collector.

sjoelund:
The C target also does not use Boehm GC... FMUs are not even linked against Boehm GC in master.

(Also the commit should not have been possible to make if the pre-commit hook was installed; the summary line in the message is too long)

sjoelund:
Note that if we did not support MSVC, we could stack-allocate this memory in most cases...

hkiel:
Is it a good idea to do this?
#ifdef __MSVC__
  // use malloc
#else
  //use stack
#endif

sjoelund:
We never use malloc. We use a memory pool. And no, it's a pretty bad idea when Francesco comes with his 10000 dimension arrays... You could also change the function to be an inline function and allocate depending on need. The memory pool is rather fast. The implementation is similar to a stack. It just does not free until the current simulation step finishes.

adrpo:
@sjoelund: it seems they are using 1.9.3 so which if I remember correctly still uses Boehm GC.
@MichaelKobierski-OPALRT: try to use the latest nightly build from: https://build.openmodelica.org/omc/builds/windows/nightly-builds/

MichaelKobierski-OPALRT:
Thank you all for your comments.  I saw the memory pool in the code in 1.9.3, but didn't realize it was eventually going to replace GC.  I will give your suggestions a try.

adrpo:
I'll close this pull request, please open a new one if needed.

PR #443, opened 2016-02-23
recursive -> loop

hkiel:
replaced some recursions by explicit loops
optimized some case SOME()/case NONE() -> case SOME()/else

### DIFF
--- a/Compiler/BackEnd/BackendDAEUtil.mo
+++ b/Compiler/BackEnd/BackendDAEUtil.mo
@@ -3462,11 +3462,11 @@ algorithm
   outM := match(inM)
   local
     BackendDAE.IncidenceMatrix m,m1;
-    case (NONE()) then NONE();
     case (SOME(m))
       equation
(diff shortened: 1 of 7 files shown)


sjoelund:
I am not sure `case SOME() case NONE()` is slower than `case SOME() else` because the former is usually optimized to a switch-statement anyway. But it does not really hurt.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-02-23_16-14-14](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1675/console).

hkiel:
I checked the generated C code and it should be faster, because a for/switch is generated and in both cases optionNone() is called if using case SOME() case NONE().

hkiel:
Also, shouldn't "else" be translated to "default:"?

sjoelund:
It should not :)

PR #446, opened 2016-02-23
use direct access to head and tail of lists

hkiel:
we already check for empty lists in the for-loop's condition, no need to do it more often.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-02-24_06-30-53](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1685/console).

PR #449, opened 2016-02-24
Fix expansion of array expressions

lochel:
This is about to fix [ticket 3716](https://trac.openmodelica.org/OpenModelica/ticket/3716).

### DIFF
--- a/Compiler/FrontEnd/Expression.mo
+++ b/Compiler/FrontEnd/Expression.mo
@@ -12173,93 +12173,70 @@ protected function traversingextendArrExp "author: Frenkel TUD 2010-07.
   output DAE.Exp outExp;
   output Boolean outExpanded;
 algorithm
-  (outExp, outExpanded) := matchcontinue(inExp, inExpanded)
+  (outExp, outExpanded) := matchcontinue(inExp)
     local
       DAE.ComponentRef cr;
-      list<DAE.ComponentRef> crlst;
-      DAE.Type t,ty;
+      DAE.Type ty;
       DAE.Dimension id, jd;
-      list<DAE.Dimension> ad;
-      Integer i,j;
-      list<list<DAE.Subscript>> subslst,subslst1;
-      list<DAE.Subscript> sublstcref;
+      Integer i, j;
       list<DAE.Exp> expl;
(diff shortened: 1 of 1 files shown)


lochel:
@adrpo Do you think this is the proper way to handle this?

lochel:
If so, I will change the other cases in the same way.

adrpo:
Unfortunately I don't know that well the array expansion parts, @mahge can you comment on this?

lochel:
Actually, I wanted to involve Mahder as well, but I couldn’t find his gh id, since he hasn’t join the OpenModelica organization yet.

adrpo:
I now invited Mahder @mahge  to join, hopefully he'll join soon.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-02-24_12-06-46](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1690/console).

mahge:
I think this should be okay. It's definitely better than what was there. 
ExpandExpression function was a quick fix for the residual function generations. Be careful with it. The problem is that if you pass a cref that is an array of records you get a list of every member X array element instead of a list of single record crefs (which you would expect from just vectorizing an array). OR if you let's say send a record cref with an array member you will also get list of every single scalar element instead of a list of just the record memebers with the array element intact. Just keep that in mind. 

There is a function traversingextendArrExp. That should give you the proper "expansions" as above.  It also is not perfect (i know...). It returns an expression. Not a list of expressions. that is something like a cref -> DAE.ARRAY(listExp) or cref -> DAE.RECORD(listExp). Actually it returns a function call to record constructor instead of a DAE.RECORD() (sigh!) and if you change that things break because the BackEnd doesn't like DAE.RECORD() ....anyway that's another story :). I have fixed most of it but not complete yet. 

So use one of the two functions based on what you need.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-02-24_15-12-36](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1692/console).

PR #452, opened 2016-02-24
Fix accidental check of uninitialized comp->loggerOn

MichaelKobierski-OPALRT:


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2016-02-25_02-04-27](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1694/).

adrpo:
This seems to break some test models, i'll investigate a bit.

sjoelund:
Check FMI2 as well. It's mostly been copy-pasted.

sjoelund:
The generated mo-files seem to set:
```
parameter Boolean debugLogging = false
```
So these debug-messages should indeed not be output.

sjoelund:
I have been chatting with @adeas31 on how to resolve this. FMIL passes the wrong loggingOn to the FMU, which means we need to call fmi1_import_set_debug_logging after `fmi1_import_instantiate_model` in our FMI import. This is the bug shown. I propose to instead remove this log-message.

PR #458, opened 2016-02-25
Test for new uniontype calling

sjoelund:
The design for new MM is that we should call the record in the union
directly instead of having an implicit import like before. This
commit tests that it is currently possible to do so in the compiler.

### DIFF
--- a/Compiler/BackEnd/BackendDAEUtil.mo
+++ b/Compiler/BackEnd/BackendDAEUtil.mo
@@ -52,64 +52,68 @@ public import DAE;
 public import FCore;
 public import Util;
 
-protected import Algorithm;
-protected import Array;
-protected import BackendDAEOptimize;
-protected import BackendDAETransform;
-protected import BackendDump;
-protected import BackendEquation;
-protected import BackendDAEEXT;
-protected import BackendInline;
-protected import BackendVarTransform;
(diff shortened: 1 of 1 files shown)


sjoelund:
Do not push this PR yet. I am just testing out functionality in OM.

sjoelund:
@lochel what do you think about changing the (entire) backend to the new MM way of calling the record constructors (DAE.Exp.ICONST, or using an import DAE.Exp; followed by Exp.ICONST)? We need to do this refactoring sometime, and could possibly start with the backend earlier since it is able to use the latest OM features.

sjoelund:
I guess for the release following 1.9.4, we will change all of the source code and disallow the original way of calling records.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-02-25_16-23-51](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1707/console).

lochel:
Seems to be a bit of work, but should be quite straight forward. Can you elaborate a bit why you want to do this changes?

lochel:
Sorry, it's quite obvious why it's a good idea to do this. So, yes, we should do it and probably start with the back end. :+1:

PR #462, opened 2016-02-26
rescursion -> loop

hkiel:


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-02-26_21-42-12](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1716/console).

hkiel:
UtilTest.mos exploded in time.
I'm just investigating why...

hkiel:
with option +d=rml the while loops are *very* slow.
I have to revert to list comprehension, which is better, but not optimal when called with integer ranges.

sjoelund:
Why not for-loops? Range expressions are optimized to not be constructed for those I am pretty sure...

hkiel:
For the list construction the items are prepended in reverse order. while seemed to be easier to write.
Also, because of the reverse construction I hesitated to re-write intRange3,because you would have to calculate the value of the last element first and then count backwards.
Any idea why using +d=rml makes things so slow? Is the while loop unrolled?

sjoelund:
No, it's just when using `-d=evalfunc` that it gets slow. It updates the environment for each assignment to the variables and it's just very slow since it needs to do lookup all the time. Not a good interpreter at all. For the while-loops they become even slower because the condition needs to be lookup+evaluated in each iteration. For-loops don't need that.

PR #463, opened 2016-02-26
resurrect list comprehensions

hkiel:
fixes PR462 (UtilTest.mos)


sjoelund:
I will change the testcase instead, adding `-d=noevalfunc` for the slow function. It is more a problem with the function interpretation than your changes.

PR #473, opened 2016-02-29
Improve performace of extractVarFromVar

wibraun:
try to fix [#3685](https://trac.openmodelica.org/OpenModelica/ticket/3685).

### DIFF
--- a/Compiler/BackEnd/BackendVariable.mo
+++ b/Compiler/BackEnd/BackendVariable.mo
@@ -812,23 +812,20 @@ public function isVarNonDiscreteAlg
 algorithm
   result := match (var)
     local
-      BackendDAE.VarKind kind;
-      BackendDAE.Type typeVar;
-      list<BackendDAE.VarKind> kind_lst;
 
     /* Real non discrete variable */
     case (BackendDAE.VAR(varType = DAE.T_REAL(_,_))) equation
-      then isVarAlg(var) or isOptInputVar(kind);
+      then (isVarAlg(var) and not isVarDiscreteRealAlg(var)) or isOptInputVar(var);
 
     else false;
   end match;
 end isVarNonDiscreteAlg;
 
-protected function isOptInputVar
-  input BackendDAE.VarKind kind;
+public function isOptInputVar
+  input BackendDAE.Var var;
   output Boolean b;
 algorithm
-  b := match(kind)
+  b := match(var.varKind)
      case BackendDAE.OPT_LOOP_INPUT()
      then true;
      case BackendDAE.OPT_INPUT_WITH_DER()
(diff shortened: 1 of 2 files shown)


OpenModelica-Hudson:
Wrong hash OMCompiler 092e0f5e260e9f75897476f4aad373ad93608066 0278ca8e3353667d659ae7e40c5f04f084be7d12

OpenModelica-Hudson:
Wrong hash OMCompiler 092e0f5e260e9f75897476f4aad373ad93608066 93b1247783751488f4d15a70eb4026717b4b53b1

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-02-29_14-16-12](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1746/console).

casella:
Compare [the new results](https://test.openmodelica.org/libraries/history/ScalableTestSuite_Experimental-2016-03-01.html) with [the old ones](https://test.openmodelica.org/libraries/history/ScalableTestSuite_Experimental-2016-02-24.html), in particular the StringModelica_N_XX models. N = 2 and N = 4 show a significant improvement, up to a factor 2. However, for larger ones the performance is still the same. I guess you have fixed a part of the code that scales badly, but there are still other one(s) that don't scale correctly, and they dominate the execution time from a certain size up.

Is it possible to have more fine-grained profiling to find out which parts of the MetaModelica code take the lion's share of the CPU time? I guess that would be very helpful...

sjoelund:
Just run valgrind on the code. It's only a 100-time slowdown :) Or perhaps use operf.

casella:
If the run time is 10 seconds, I guess Willi can afford a 100x slowdown.

wibraun:
I'm quite confident that creating all SimVars is dominating the execution time in the SimCode phase, at least in all models I have tested. 
This job is done by the function extractVarFromVar, which basically creates a SimVar from a BackendVar and select the correct list for the corresponding variable. Anyway, here I've just tried to optimize this act, I guess we need here a better approach for handling of a lot long lists.

casella:
But why creating two times as many variables should take 3.5 times more time? There really must be something fundamentally wrong in how these data structures are managed.

wibraun:
Okay, I was wrong it's the function preCalculateStartValues, which takes so long.

hkiel:
Valgrind exposes a huge amount of work in GC (at least here on my OSX machine).
I saw about 1/3rd of the time spent in GC, not the model.
Also -d=gcProfiling shows that 16GB or RAM were allocated (and "freed") during a simple simulation run.
I propose investigation of the use of #481 to get better results.
I suspect listAppend() is one of the big culprits.

sjoelund:
```mo
rest := listAppend(rest,newEqIdcs); // Appending to the end of the list is slow :(
```

See #481 for an explanation of why

sjoelund:
Actually, it seems the result of that iteration is not even used. So it can probably be removed (and it scales O(n^2))...

sjoelund:
Never mind, it's used for some weird iteration building a list... Really odd case.

PR #481, opened 2016-03-01
Implement MetaModelica.Dangerous.listAppendTail()

hkiel:
it is a fast and not memory consuming alternative to construct lists using "listAppend" method.
Sample implementation of some functions given.

Should be able to improve [ticket 3685](https://trac.openmodelica.org/OpenModelica/ticket/3685)

### DIFF
--- a/SimulationRuntime/c/meta/meta_modelica_builtin.c
+++ b/SimulationRuntime/c/meta/meta_modelica_builtin.c
@@ -443,6 +443,61 @@ modelica_metatype listAppend(modelica_metatype lst1,modelica_metatype lst2)
   return MMC_TAGPTR(res);
 }
 
+modelica_metatype listAppendTail(modelica_metatype lst1, modelica_metatype tail1, modelica_metatype lst2, modelica_metatype *outTail)
+{
+  struct mmc_cons_struct *res = NULL;
+  struct mmc_cons_struct *p = NULL;
+  modelica_metatype tail2;
+  int length;
+  int i;
+  modelica_metatype head = MMC_UNTAGPTR(lst1);
+
+  if (MMC_NILTEST(lst2)) { /* If lst2 is empty, simply return lst1 and old tail of lst1*/
+    *outTail = tail1;
+    return lst1;
+  }
+  if (MMC_NILTEST(lst1)) { /* If lst1 is empty, simply return lst1 and old tail of empty lst1 */
+    *outTail = tail1;
+    return lst2;
+  }
+  length = listLength(lst2);
+  if (MMC_NILTEST(tail1)) {
+    // copy list one if not already done for this instance
+    int length1 = listLength(lst1);
+    res = (struct mmc_cons_struct*)mmc_alloc_words( length1+length * 3 ); /* Do one single big alloc. It's cheaper */
+    head = res;
+    for (i=0; i<length1; i++) { /* Write all except the last element... */
+      p = res+i;
+      p->header = MMC_STRUCTHDR(2, MMC_CONS_CTOR);
+      p->data[0] = MMC_CAR(lst1);
+      p->data[1] = MMC_TAGPTR(res+i+1);
+      lst1 = MMC_CDR(lst1);
+    }
+    res = res + i;
+  } else {
+    res = (struct mmc_cons_struct*)mmc_alloc_words( length * 3 ); /* Do one single big alloc. It's cheaper */
+    p = MMC_UNTAGPTR(tail1);
+    p->data[1] = MMC_TAGPTR(res);
+  }
+
+  // copy list two
+  for (i=0; i<length-1; i++) {
+    p = res+i;
+    p->header = MMC_STRUCTHDR(2, MMC_CONS_CTOR);
+    p->data[0] = MMC_CAR(lst2);
+    p->data[1] = MMC_TAGPTR(res+i+1);
+    lst2 = MMC_CDR(lst2);
+  }
+  // last element terminates the list
+  p = res+i;
+  p->header = MMC_STRUCTHDR(2, MMC_CONS_CTOR);
+  p->data[0] = MMC_CAR(lst2);
+  p->data[1] = MMC_CDR(lst2);
+
+  *outTail = MMC_TAGPTR(p);
+  return MMC_TAGPTR(head);
+}
+
 modelica_integer listLength(modelica_metatype lst)
 {
   modelica_integer res = 0;
--- a/Compiler/BackEnd/BackendVariable.mo
+++ b/Compiler/BackEnd/BackendVariable.mo
@@ -2044,20 +2044,17 @@ end listVar1;
 public function equationSystemsVarsLst
   input BackendDAE.EqSystems systs;
   input list<BackendDAE.Var> inVars;
-  output list<BackendDAE.Var> outVars;
+  output list<BackendDAE.Var> outVars = {};
+protected
+  list<BackendDAE.Var> tail = {};
+  BackendDAE.Variables v;
+  list<BackendDAE.Var> vars;
 algorithm
-  outVars := match (systs)
-    local
-      BackendDAE.EqSystems rest;
-      list<BackendDAE.Var> vars, vars1;
-      BackendDAE.Variables v;
-
-    case {} then inVars;
-    case BackendDAE.EQSYSTEM(orderedVars=v)::rest equation
-      vars = varList(v);
-      vars1 = listAppend(inVars,vars);
-    then equationSystemsVarsLst(rest,vars1);
-  end match;
+  for vs in inVars loop
+    BackendDAE.EQSYSTEM(orderedVars=v) := vs;
+    vars := varList(v);
+    (vars1,tail) := listAppendTail(inVars,tail,vars);
+  end for;
 end equationSystemsVarsLst;
 
 public function daeVars "returns orderedVars"
--- a/Compiler/BackEnd/EvaluateFunctions.mo
+++ b/Compiler/BackEnd/EvaluateFunctions.mo
@@ -56,6 +56,7 @@ protected import List;
 protected import RemoveSimpleEquations;
 protected import SCode;
 protected import Util;
+protected import MetaModelica.Dangerous;
 
 
 // =============================================================================
@@ -2620,7 +2621,7 @@ algorithm
       DAE.Subscript sub;
       list<DAE.Dimension> rest;
       list<DAE.Subscript> subs;
-      list<list<DAE.Subscript>> subsLst, subsLst1, subFold = {};
+      list<list<DAE.Subscript>> subsLst, subsLst1, subFold = {}, tail = {};
   case(dim::rest,_)
     algorithm
       size := Expression.dimensionSize(dim);
@@ -2629,7 +2630,7 @@ algorithm
       subsLst := List.map(subs,List.create);
       for sub in subsIn loop
         subsLst1 := List.map1r(subsLst,listAppend,sub);
-        subFold := listAppend(subFold,subsLst1);
+        (subFold, tail) := Dangerous.listAppendTail(subFold,{},subsLst1);
       end for;
       if listEmpty(subsIn) then subFold := subsLst; end if;
     then expandDimension(rest,subFold);
--- a/Compiler/FrontEnd/MetaModelicaBuiltin.mo
+++ b/Compiler/FrontEnd/MetaModelicaBuiltin.mo
@@ -901,6 +901,14 @@ function listReverseInPlace<A> "O(n). A destructive listReverse. May cause segme
 external "builtin";
 end listReverseInPlace;
 
+function listAppendTail<A> "O(length(lst1)), O(1) if either list is empty"
+  input List<A> lst1;
+  input List<A> tail;
+  input List<A> lst2 = {};
+  output List<A> lst;
+  output List<A> outTail;
+external "builtin";
+end listAppendTail;
 end Dangerous;
 
 end MetaModelica;
--- a/SimulationRuntime/c/meta/meta_modelica_builtin.h
+++ b/SimulationRuntime/c/meta/meta_modelica_builtin.h
@@ -98,6 +98,7 @@ extern modelica_metatype listReverse(modelica_metatype);
 extern modelica_metatype listReverseInPlace(modelica_metatype);
 extern modelica_boolean listMember(modelica_metatype, modelica_metatype);
 extern modelica_metatype listAppend(modelica_metatype,modelica_metatype);
+extern modelica_metatype listAppendTail(modelica_metatype,modelica_metatype,modelica_metatype,modelica_metatype *);
 extern modelica_integer listLength(modelica_metatype);
 #define listGet(X,Y) boxptr_listGet(threadData,X,mmc_mk_icon(Y))
 #define listEmpty(LST) MMC_NILTEST(LST)


sjoelund:
This PR seems wrong since it allocates as much memory as before. The point is to allocate less memory. This is how to do it in zero memory:
```mo
if listEmpty(l1) then
  res:=l2;
elseif listEmpty(l2) then
  res:=l1;
else
  setCDR(List.last(l1), l2); // Yes, just update the reference here
  res:=l1;
end if;
```

The effect would be that you don't need to listReverse things in order to make tail-recursive listAppend things, which saves memory and a lot of time. It still takes linear time to find the last element though (which means quadratic in the poor algorithms used above).

The MetaModelica prototype seems correct, but builtin functions do not return two values. Storing the tail does reduce the cost of the function, and does make it simpler to iterate...

Old bad style in loops:
```mo
lst := {};
for x in xs loop
  lst := listAppend(lst, x); // O(n^2)
end for;
```
Old good style in loops:
```mo
lst := {};
for x in xs loop
  lst := listAppend(listReverse(x), lst); // 2*n
end for;
lst := listReverseInPlace(lst); // n
// 3*n operations
```
This proposal:
```mo
lst := {};
tail := {};
for x in xs loop
  (lst,tail) := listAppendTail(lst, x, tail); // 1
end for;
// listLength(xs) operations
```
Other alternative:
```mo
lst := {};
for x in xs loop
  lst := listAppendInPlace(lst, x); // O(n^2) operations, but only O(n) allocations
end for;
```

wibraun:
There seems be be something fishy, since I run into "Segmentation fault" or "Floating point exception"  issues.

sjoelund:
Yes, the C-code for the function is really off

hkiel:
C code was "borrowed" from listAppend. I'm not too familiar with MetaModelica pointer handling (MMC_TAGPTR() etc.) and did my best.

My concern (when just adjusting pointer of last element in list 1) was that using listAppendInPlace would destroy also the second list if listAppendInPlace is called again (appends to the "middle" list which might be used elsewhere).

BTW, my test code ran very well (producing correct output) and rather quick and even using much less memory than old bad style loop.

Anyway, wouldn't "old good style loop" need another xs:=listReverse(xs) beforehand?

This is my test code. For large parameters n and m I just commented out r2.
```
function test_listAppend
  input Integer n;
  input Integer m;
  output Integer r = n*m;
  output list<Integer> r2;
protected
  list<Integer> l;
  list<list<Integer>> ll = {};
algorithm
  for i in 1:n loop
    l := list(j for j in i:(i+m));
    ll := l :: ll;
  end for;

  l := {};
  for k in ll loop
    l := listAppend(l,k);
  end for;
  r2 := l;
end test_listAppend;

function test_listAppendTail
  input Integer n;
  input Integer m;
  output Integer r = n*m;
  output list<Integer> r2;
protected
  list<Integer> l;
  list<Integer> tail = {};
  list<list<Integer>> ll = {};
algorithm
  for i in 1:n loop
    l := list(j for j in i:(i+m));
    ll := l :: ll;
  end for;

  l := {};
  for k in ll loop
    (l,tail) := MetaModelica.Dangerous.listAppendTail(l,tail,k);
  end for;
  r2 := l;
end test_listAppendTail;
```

sjoelund:
Nope, the old good style only requires the one listReverseInPlace at the end; List.append_reverse does the only necessary memory allocation. It really is quite fast and does not require as many temporary variables as `listAppendTail` does. Anyway, I would implement it slightly differently to be pure performance assuming people don't share lists (probably by adding `Dangerous.setRest()` or `setCDR` and implementing the rest as a regular MM function).

SimulationRuntime/c/meta/meta_modelica_builtin.c
```diff
@@ -443,6 +443,61 @@ modelica_metatype listAppend(modelica_metatype lst1,modelica_metatype lst2)
...
+  if (MMC_NILTEST(lst2)) { /* If lst2 is empty, simply return lst1 and old tail of lst1*/
+    *outTail = tail1;
+    return lst1;
+  }
+  if (MMC_NILTEST(lst1)) { /* If lst1 is empty, simply return lst1 and old tail of empty lst1 */
+    *outTail = tail1;
+    return lst2;
+  }
+  length = listLength(lst2);
+  if (MMC_NILTEST(tail1)) {
```
sjoelund: This part is unreachable code.

SimulationRuntime/c/meta/meta_modelica_builtin.c
```diff
@@ -443,6 +443,61 @@ modelica_metatype listAppend(modelica_metatype lst1,modelica_metatype lst2)
...
+    }
+    res = res + i;
+  } else {
+    res = (struct mmc_cons_struct*)mmc_alloc_words( length * 3 ); /* Do one single big alloc. It's cheaper */
+    p = MMC_UNTAGPTR(tail1);
+    p->data[1] = MMC_TAGPTR(res);
+  }
+
+  // copy list two
+  for (i=0; i<length-1; i++) {
```
sjoelund: Copying should be done manually in a dangerous function. Most use-cases won't need this copy

hkiel:
listAppend() also does this copying for the first list. So, that would be superfluous then?

hkiel:
Thanks @wibraun for the fix. I wonder why it did not produce a compilation error here.

This is my test scenario.

12000 MB using listAppend
32 MB using listAppendTail
112 MB using append_reverse


ListPerf.mo:
```
function test_append_reverse
  input Integer n;
  input Integer m;
  output Integer r;
algorithm
  (r,_) := rtest_append_reverse(n,m);
end test_append_reverse;

function test_listAppendTail
  input Integer n;
  input Integer m;
  output Integer r;
algorithm
  (r,_) := rtest_listAppendTail(n,m);
end test_listAppendTail;

function test_listAppend
  input Integer n;
  input Integer m;
  output Integer r;
algorithm
  (r,_) := rtest_listAppend(n,m);
end test_listAppend;

function rtest_listAppend
  input Integer n;
  input Integer m;
  output Integer r = n*m;
  output list<Integer> r2;
protected
  list<Integer> l;
  list<list<Integer>> ll = {};
algorithm
  for i in 1:n loop
    l := list(j for j in i:(i+m));
    ll := l :: ll;
  end for;
  l := {};
  for k in ll loop
    l := listAppend(l,k);
  end for;
  r2 := l;
end rtest_listAppend;

function rtest_listAppendTail
  input Integer n;
  input Integer m;
  output Integer r = n*m;
  output list<Integer> r2;
protected
  list<Integer> l;
  list<Integer> tail = {};
  list<list<Integer>> ll = {};
algorithm
  for i in 1:n loop
    l := list(j for j in i:(i+m));
    ll := l :: ll;
  end for;
  l := {};
  for k in ll loop
    (l,tail) := MetaModelica.Dangerous.listAppendTail(l,tail,k);
  end for;
  r2 := l;
end rtest_listAppendTail;

function append_reverse<T>
  "Appends the elements from list1 in reverse order to list2."
  input list<T> inList1;
  input list<T> inList2;
  output list<T> outList=inList2;
algorithm
  if listEmpty(outList) then
    outList := listReverse(inList1);
    return;
  end if;
  for e in inList1 loop
    outList := e::outList;
  end for;
end append_reverse;

function rtest_append_reverse
  input Integer n;
  input Integer m;
  output Integer r = n*m;
  output list<Integer> r2;
protected
  list<Integer> l;
  list<list<Integer>> ll = {};
algorithm
  for i in 1:n loop
    l := list(j for j in i:(i+m));
    ll := l :: ll;
  end for;
  l := {};
  for k in ll loop
    l := append_reverse(k,l);
  end for;
  //l := MetaModelica.Dangerous.listReverseInPlace(l);
  l := listReverse(l);
  r2 := l;
end rtest_append_reverse;
```

ListPerf.mos: // shows same output for all three variants
```
loadFile("ListPerf.mo");
rtest_listAppend(3,3);
rtest_listAppendTail(3,3);
rtest_append_reverse(3,3);
````

ListPerf1.mos: // bad old style listAppend
```
loadFile("ListPerf.mo");
test_listAppend(1000,1000);
````

ListPerf2.mos: // using my listAppendTail()
```
loadFile("ListPerf.mo");
test_listAppendTail(1000,1000);
````

ListPerf3.mos: // new append_reverse style
```
loadFile("ListPerf.mo");
test_append_reverse(1000,1000);
````

```
> time omc -g=MetaModelica -d=rml,noevalfunc,gcProfiling ListPerf.mos ; for i in 1 2 3 ; do time omc -g=MetaModelica -d=rml,noevalfunc,gcProfiling ListPerf$i.mos ; done
GC stats after initialization:
  heapsize_full: 262144
  free_bytes_full: 221184
  unmapped_bytes: 0
  bytes_allocd_since_gc: 11744
  allocd_bytes_before_gc: 0
  non_gc_bytes: 0
  gc_no: 1
  markers_m1: 3
  bytes_reclaimed_since_gc: 0
  reclaimed_bytes_before_gc: 0
true
(9,{3,4,5,6,2,3,4,5,1,2,3,4})
(9,{3,4,5,6,2,3,4,5,1,2,3,4})
(9,{3,4,5,6,2,3,4,5,1,2,3,4})
GC stats at end of program:
  heapsize_full: 7950336
  free_bytes_full: 2764800
  unmapped_bytes: 0
  bytes_allocd_since_gc: 582352
  allocd_bytes_before_gc: 11070640
  non_gc_bytes: 0
  gc_no: 7
  markers_m1: 3
  bytes_reclaimed_since_gc: 3074336
  reclaimed_bytes_before_gc: 3513728

real    0m1.174s
user    0m0.735s
sys     0m0.279s
GC stats after initialization:
  heapsize_full: 262144
  free_bytes_full: 221184
  unmapped_bytes: 0
  bytes_allocd_since_gc: 11744
  allocd_bytes_before_gc: 0
  non_gc_bytes: 0
  gc_no: 1
  markers_m1: 3
  bytes_reclaimed_since_gc: 0
  reclaimed_bytes_before_gc: 0
true
1000000
GC stats at end of program:
  heapsize_full: 192552960
  free_bytes_full: 108212224
  unmapped_bytes: 0
  bytes_allocd_since_gc: 23999984
  allocd_bytes_before_gc: 12023845056
  non_gc_bytes: 0
  gc_no: 205
  markers_m1: 3
  bytes_reclaimed_since_gc: 71783744
  reclaimed_bytes_before_gc: 11891785008

real    0m7.313s
user    0m11.816s
sys     0m0.278s
GC stats after initialization:
  heapsize_full: 262144
  free_bytes_full: 221184
  unmapped_bytes: 0
  bytes_allocd_since_gc: 11744
  allocd_bytes_before_gc: 0
  non_gc_bytes: 0
  gc_no: 1
  markers_m1: 3
  bytes_reclaimed_since_gc: 0
  reclaimed_bytes_before_gc: 0
true
1000000
GC stats at end of program:
  heapsize_full: 59551744
  free_bytes_full: 20086784
  unmapped_bytes: 0
  bytes_allocd_since_gc: 2958400
  allocd_bytes_before_gc: 69071664
  non_gc_bytes: 0
  gc_no: 11
  markers_m1: 3
  bytes_reclaimed_since_gc: 22570864
  reclaimed_bytes_before_gc: 10137600

real    0m0.682s
user    0m0.633s
sys     0m0.134s
GC stats after initialization:
  heapsize_full: 262144
  free_bytes_full: 221184
  unmapped_bytes: 0
  bytes_allocd_since_gc: 11744
  allocd_bytes_before_gc: 0
  non_gc_bytes: 0
  gc_no: 1
  markers_m1: 3
  bytes_reclaimed_since_gc: 0
  reclaimed_bytes_before_gc: 0
true
1000000
GC stats at end of program:
  heapsize_full: 109867008
  free_bytes_full: 9392128
  unmapped_bytes: 0
  bytes_allocd_since_gc: 44318720
  allocd_bytes_before_gc: 68280400
  non_gc_bytes: 0
  gc_no: 11
  markers_m1: 3
  bytes_reclaimed_since_gc: 1575232
  reclaimed_bytes_before_gc: 10623392

real    0m0.607s
user    0m0.449s
sys     0m0.150s
````

**Comment by sjoelund** [comment]
This should still be a user library and not a MM builtin I think. I'll cook up something nicer

**Comment by sjoelund** [comment]
I get your numbers to:
2: 72MB
3: 112MB
3 with listReverseInPlace: 84MB (measured by myself)

So 2 and 3 are really similar in performance. I think just adding MetaModelica.Dangerous.listSetRest would be enough to implement most of these kind of functions:
```
function listSetRest<A> "O(1). A destructive operation changing the \"rest\" part of a cons-cell.
NOTE: Make sure you do NOT create cycles as infinite lists are not handled well in the compiler."
  input list<A> inConsCell "A non-empty list";
  input list<A> inNewRest;
external "builtin";
end listSetRest;
```

**Comment by sjoelund** [comment]
Removing the overhead for compiling the function:
2 with listSetRest: 40MB
3 with listReverseInPlace: 72MB

The main difference in performance here is that (2) assumes you do not need to copy the input list (so it works for the given list reduction, which will allocate memory; 3 does a List.append_reverse, which copies the data and is thus "safer" or handles more cases).

```mo
function lastCell<X>
  "Returns the last element of a list. Fails if the list is empty."
  input list<X> inList;
  output list<X> e=inList;
protected
  list<X> rest=listRest(inList);
algorithm
  while not listEmpty(rest) loop
    e as (_::rest) := rest;
  end while;
end lastCell;

function listAppendTail<A> "O(length(lst1)), O(1) if either list is empty"
  input list<A> lst1;
  input list<A> lst2;
  input list<A> tail = if listEmpty(lst1) then {} else lastCell(lst1);
  output list<A> lst "lst1 appended to lst2 by a destructive fuse of the last element of lst1 (which may be passed explicitly in order to improve performance)";
  output list<A> outTail "The last element of lst";
algorithm
  if listEmpty(lst1) then
    lst := lst2;
    outTail := lastCell(lst2);
    return;
  elseif listEmpty(lst2) then
    lst := lst1;
    outTail := tail;
    return;
  else
    false := listEmpty(tail); // Or an internal error
    MetaModelica.Dangerous.listSetRest(tail, lst2);
    lst := lst1;
    outTail := lastCell(lst2);
  end if;
end listAppendTail;
```

hkiel:
Sounds like a good solution to me.

Just to summarize:
- my patch is not needed, because if you know what you do you can even spare more bytes using your new proposed way
- GC is an issue for large(r) models
- listAppend() as a single function is rather optimal, but can become far from optimal when used recursively -> use loop + append_reverse() instead
- we should keep an eye on "mis-use" of listAppend()

sjoelund:
I agree listAppend mis-use should be reduced. I forgot to write it yesterday, but I would propose making a compiler warning for the following pattern:
```mo
x = listAppend(x, _);
```

lochel:
That’s a good idea! Maybe it’s enough to trigger a warning if such a statement is used within a loop. But it is probably difficult to detect recursions, so a warning for all these expressions is probably the better approach.

sjoelund:
It is probably not too hard to detect if it is used within a loop. However, there are many instances where people iterate without a for-loop. Such as appending initial variables, discrete variables, parameter variables, etc. If it is done in the wrong order, you get O(N^2) scaling also without loops.

casella:
Please  do. It is important that we catch these programming patterns pre-emptively.

sjoelund:
I will make it a notification instead of a warning at first (because warnings become errors when bootstrapping and there are too many warnings to fix them all for me).

sjoelund:
I will also add an annotation `__OpenModelica_DisableListAppendWarning=true` in case we have some weird case where we know doing the wrong order `listAppend` is faster (I guess usually only true if there is exactly two lists to append and side-effects when generating the first list).

sjoelund:
See PR #486 about that fix. There are 634 potentially bad listAppend occurances in the code.

casella:
How long would it take to fix them all?

sjoelund:
A long time unless I get some help. Got more pressing matters.

PR #484, opened 2016-03-02
Add support for MA Impure annotation

sjoelund:
This annotation is handled like the impure prefix, and not like
__OpenModelica_Impure=true (which also makes the function unavailable
for parameters).


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-03-02_07-01-19](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1765/console).

PR #486, opened 2016-03-02
Notify the user about listAppend in wrong order

sjoelund:


OpenModelica-Hudson:
The job failed the coding style; for details, see [OpenModelica_TEST_PULL_REQUEST 2016-03-02_11-42-05](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1768/console). Remember to clean up the commit you did that breaks the coding style rather than piling a new commit on top of the old ones. This goes double for accidental addition of files that you need to remove.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-03-02_12-06-59](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1770/console).

casella:
I guess the correct description of this PR should have been "Notify the _developers_ about listAppend in wrong order. I don't think the average OMC user will have the foggiest clue of what the hell "listAppend in wrong order" means :)

sjoelund:
The developers are my users

casella:
My point is: these warnings will also be read by end-users, which will be totally confused. Let's not forget that we are working for them!
I think it would be better to have these notifications optionally activated by a debug flag with default = false

adrpo:
What @sjoelund is trying to say is that these messages will never be seen by Modelica users, just by the compiler developers. There is no listAppend in Modelica (at least not yet).

casella:
Oops, sorry, I understood that these messages would have popped out each time a list was built in the wrong way during the processing of end user's models.

Just forget my stupid commment...

adrpo:
No worries @casella, @sjoelund is a bit cryptic sometimes :)

casella:
I'd rather say terse, or concise :)

PR #491, opened 2016-03-02
Remove FMI1 log-message

sjoelund:
The reason for removing this log-message is that FMIL passes loggingOn
as true even if we call instantiate with loggingOn=false. The FMU would
work correctly when FMIL was not used, but the OM FMU import does use
FMIL and it is annoying to see the GUID when not debugging.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-03-02_14-56-19](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1779/console).

PR #494, opened 2016-03-02
optimize some list operations

hkiel:
matchcontinue -> match


Compiler/BackEnd/Differentiate.mo
```diff
@@ -2655,15 +2655,15 @@ algorithm
     String matrixName;
 
     case ({}, _, _, _, _, _, _, _)
-    then (inElementsDer, inFunctionTree, inElementsNoDer, inBooleanLst);
+    then (MetaModelica.Dangerous.listReverseInPlace(inElementsDer), inFunctionTree, MetaModelica.Dangerous.listReverseInPlace(inElementsNoDer), MetaModelica.Dangerous.listReverseInPlace(inBooleanLst));
```
sjoelund: listReverseInPlace does not belong here: the function can be called with a non-empty input which is messed up.

hkiel:
It *can* be, but isn't.
Also it is a protected function, so it should be visible from the module if it is used wrong.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-03-03_08-32-48](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1792/console).

PR #500, opened 2016-03-03
Reduce memory usage

hkiel:


sjoelund:
`listReverseArray` should not be a built-in function, but rather `Array.fromReverseList` or `List.toReverseArray`. It is very simple to implement it efficiently using `arrayCreateNoInit` or by `Array.reverse(listArray(lst))` (slower, but faster than `listReverse` first).

PR #501, opened 2016-03-03
try to reduce memory usage

hkiel:
- use listReverseInPlace where possible
- use new listReverseArray(l) instead of listArray(listReverse(l))
- remove some recursions and matchcontinue
- avoid List.intersection1OnTrue if only one of the outputs is needed

### DIFF
--- a/Compiler/BackEnd/BackendDAEOptimize.mo
+++ b/Compiler/BackEnd/BackendDAEOptimize.mo
@@ -102,7 +102,7 @@ algorithm
       else eq::removedEqsList;
     end match;
   end for;
-  shared.removedEqs := BackendEquation.listEquation(listReverse(removedEqsList));
+  shared.removedEqs := BackendEquation.listEquation(MetaModelica.Dangerous.listReverseInPlace(removedEqsList));
   outDAE.shared := shared;
 end simplifyAllExpressions;
 
@@ -4291,7 +4291,7 @@ algorithm
             end if;
             expLst := e :: expLst;
           end for; // lhs
-          left := DAE.TUPLE(listReverse(expLst));
+          left := DAE.TUPLE(MetaModelica.Dangerous.listReverseInPlace(expLst));
           eqn := BackendDAE.COMPLEX_EQUATION(size, left, right, source, attr);
           eqns := BackendEquation.setAtIndex(eqns, i, eqn);
         end if; // lhs <-> rhs
--- a/Compiler/BackEnd/BackendDAETransform.mo
+++ b/Compiler/BackEnd/BackendDAETransform.mo
@@ -161,7 +161,7 @@ algorithm
     outComps := acomp :: outComps;
   end for;
 
-  outComps := listReverse(outComps);
+  outComps := MetaModelica.Dangerous.listReverseInPlace(outComps);
 end analyseStrongComponentsScalar;
 
 protected function analyseStrongComponentScalar "author: Frenkel TUD 2011-05"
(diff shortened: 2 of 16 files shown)


sjoelund:
Can you `git commit --amend` + `git push --force` to remove mention of `listReverseArray`?

hkiel:
Done.

sjoelund:
Thanks. Hudson started, so let's hope it passes. I really appreciate your fixes.

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2016-03-03_19-41-43](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1796/).

hkiel:
It already encountered a few Failures. I will need to investigate what is going wrong.
Maybe some matchcontinue/match changes...
The models I checked locally all worked fine.

hkiel:
I found two places (one confused true/false, and one place where listReverseInPlace was wrong).
At least EngineV6 works again here. Can you please start hudson, to see what is still not working?

casella:
@hkiel, just ask a password for hudson to @adrpo or @sjoelund

hkiel:
I already did with negative response.
But I still think it is good to have the code reviewed by someone else before it can be pushed to the repo.

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2016-03-04_10-56-20](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1799/).

hkiel:
Question @sjoelund @lochel @adrpo:
List.setDifferenceOnTrue() removes only the first occurrence of each of the items to remove. Is that correct?

I would expect that function to throw out *all* occurences of the items to remove. Or is that another function?

sjoelund:
It's more like sets are assumed to not contain duplicates...

sjoelund:
Also, one thing I would like to implement in MetaModelica is the old `isPresent()` operator that was removed from the Modelica language. It is actually not hard to implement for arguments `2..N` and could mean we do not generate lists that are not used by the caller...

hkiel:
Ok, but then I would expect List.intersection1OnTrue() to behave like working on sets as well.
Maybe everything is working fine with true sets, but one is more robust than the other...

hkiel:
Is it okay to write a new function
List.removeOnTrue() with the same signature, but removing *all* occurences (i.e. not assuming uniqueness)?

sjoelund:
Isn't that just `List.filter`?

perost:
There already is a List.removeOnTrue.

sjoelund:
Or something like `list(e for e guard not List.exists(e,lst2,func) in lst1)`

hkiel:
blind me

hkiel:
I removed the changes to the sets, now it works locally. Can you start hudson again, please?

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-03-04_15-02-00](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1804/console).

sjoelund:
In #507 I try to show how to use `isPresent`. I am unsure if the compiler allows it to be used in the frontend yet though. For example things like `List.splitOnBoolList` were used in a way that the second output parameter was never used... So now we can check for this and make conditional memory allocations.

PR #515, opened 2016-03-07
SM workaround removed for Cpp runtime

bernhard-thiele:
If models are translated for the Cpp runtime the synchronous features workaround for state machines is not activated.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-03-07_09-59-30](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1837/console).

lochel:
What is missing in terms of the c runtime?

bernhard-thiele:
Just realized that things have improved for the C runtime, too :-)
Tests currently won't work, because there is no default clock and if not providing a Clock the execution will hang. However, if I provide a clock I can execute state machines. Hence, it seems the workaround can be removed soon.

lochel:
It sounds reasonable to me to just introduce a default clock if there would be none otherwise.

PR #522, opened 2016-03-10
Handle lost information for MetaModelica record

sjoelund:
This fixes OM ticket #3757 by checking knownSingleton=true on either
lhs or rhs instead of only on the rhs (which loses information after
passing through for example Ceval or Expression.typeof).


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-03-10_12-44-08](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1864/console).

PR #524, opened 2016-03-10
Reorder codegen for patterns

sjoelund:
Calculate patterns that do structural checks before patterns that do
not perform any structural checks. (So deconstructing a list is
performed before a simple assignment, etc).

Testing performance after idea from @hkiel

### DIFF
--- a/Compiler/FrontEnd/Patternm.mo
+++ b/Compiler/FrontEnd/Patternm.mo
@@ -716,7 +716,9 @@ algorithm
   end matchcontinue;
 end patternStr;
 
-public function elabMatchExpression
+public
+
+function elabMatchExpression
   input FCore.Cache inCache;
   input FCore.Graph inEnv;
   input Absyn.Exp matchExp;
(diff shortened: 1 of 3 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-03-11_06-36-48](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1868/console).

sjoelund:
Hard to tell if this gives any performance improvement. Perhaps it was better before since it was sequential array accesses that were performed?

hkiel:
Hard to tell, because largest overhead currently comes from garbage collection and timing there is unpredictable.

As I already said, it might be that the speed advantage could already be done by C compiler, or maybe invalid because of out-of-order execution of CPU.
At least there is no major advantage, so better go for simplicity...

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-03-11_12-21-34](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1871/console).

sjoelund:
Ran it again, results were even worse... It could have something to do with the fact that we use our own tmp-array and re-use memory locations in it so that clang/gcc get confused and do not optimize the code better.

hkiel:
I doubt that this minor increase in time was due to your changes.
Given that the test incorporates running clang and a lot of garbage collections, it might well be that by chance it ran slower.
Anyway, it does not seem to increase speed, so better leave it as it is.

PR #529, opened 2016-03-15
Introduce flags to handle huge systems efficiently

ptaeuber:
Default is a maximum system size of 10000 for now. Any other suggestions?

### DIFF
--- a/Compiler/BackEnd/Tearing.mo
+++ b/Compiler/BackEnd/Tearing.mo
@@ -249,7 +249,7 @@ algorithm
       equality(jacType = BackendDAE.JAC_LINEAR());
       maxSize = Flags.getConfigInt(Flags.MAX_SIZE_LINEAR_TEARING);
       if intGt(listLength(vindx),maxSize) then
-        Error.addMessage(Error.MAX_TEARING_SIZE, {intString(listLength(vindx)),"Linear",intString(maxSize)});
+        Error.addMessage(Error.MAX_TEARING_SIZE, {intString(listLength(vindx)),"linear",intString(maxSize)});
         fail();
       end if;
       if Flags.isSet(Flags.TEARING_DUMP) or Flags.isSet(Flags.TEARING_DUMPVERBOSE) then
@@ -273,7 +273,7 @@ algorithm
       failure(equality(jacType = BackendDAE.JAC_LINEAR()));
       maxSize = Flags.getConfigInt(Flags.MAX_SIZE_NONLINEAR_TEARING);
       if intGt(listLength(vindx),maxSize) then
-        Error.addMessage(Error.MAX_TEARING_SIZE, {intString(listLength(vindx)),"Nonlinear",intString(maxSize)});
+        Error.addMessage(Error.MAX_TEARING_SIZE, {intString(listLength(vindx)),"nonlinear",intString(maxSize)});
         fail();
       end if;
       if Flags.isSet(Flags.TEARING_DUMP) or Flags.isSet(Flags.TEARING_DUMPVERBOSE) then
--- a/Compiler/BackEnd/Tearing.mo
+++ b/Compiler/BackEnd/Tearing.mo
@@ -102,7 +102,7 @@ algorithm
       true = stringEqual(methodString, "noTearing");
     then inDAE;
 
-    // get method function and traveres systems
+    // get method function and traverse systems
     case(_) equation
       methodString = Config.getTearingMethod();
       BackendDAE.SHARED(backendDAEType=DAEtype) = inDAE.shared;
@@ -236,6 +236,7 @@ protected
 algorithm
   (oComp, outRunMatching) := matchcontinue (inComp, isyst, ishared, inMethod)
     local
+      Integer maxSize;
       list<Integer> eindex, vindx;
       Boolean b, b1;
       BackendDAE.StrongComponents comps, acc;
@@ -246,6 +247,11 @@ algorithm
 
     case ((BackendDAE.EQUATIONSYSTEM(eqns=eindex, vars=vindx, jac=BackendDAE.FULL_JACOBIAN(ojac), jacType=jacType, mixedSystem=mixedSystem)), _, _, _) equation
       equality(jacType = BackendDAE.JAC_LINEAR());
+      maxSize = Flags.getConfigInt(Flags.MAX_SIZE_LINEAR_TEARING);
+      if intGt(listLength(vindx),maxSize) then
+        Error.addMessage(Error.MAX_TEARING_SIZE, {intString(listLength(vindx)),"Linear",intString(maxSize)});
+        fail();
+      end if;
       if Flags.isSet(Flags.TEARING_DUMP) or Flags.isSet(Flags.TEARING_DUMPVERBOSE) then
         print("\nCase linear in traverseComponents\nUse Flag '+d=tearingdumpV' for more details\n\n");
       end if;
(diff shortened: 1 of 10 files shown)


vruge:
How about -1, where -1 represent Inf.

maxSize < 0 or intGt(listLength(vindx),maxSize)

sjoelund:
The `print()` statements are totally out of place. It should be an Error.mo notification. The limit should probably be a lot lower as well, at least for linear systems where it makes sense to try a sparse solver instead, right?

casella:
A few comments:
- @vruge, I would definitely put a reasonable finite limit as default, infinity is not a good choice for inexperienced users
- @ptaeuber I would suggest to introduce different limits for linear and nonlinear systems. For nonlinear systems, tearing can enhance convergence, so one might want to keep a larger limit; for linear systems, I understand this is mostly a matter of performance
- @ptaeuber I'm not sure if 10000 is a good limit, I would suggest to run the ScalableTestSuite examples with large linear systems (e.g., DistributionSystem*) and investigate the trade off between additional compile time and relative change of simulation time. Can you do that and report the results?
- @ptaeuber it is also essential to introduce corresponding defaults (one for linear, one for nonlinear) for switching to a sparse solver in the runtime, as suggested by @sjoelund elsewhere - otherwise, if tearing is not performed and a dense solver is used, the performance will suck. We can use klu as default sparse linear solver, and in the future add a sparse nonlinear solver (kinsol, I guess)

vruge:
@casella I like inf for defaulte user more, because the defaulte user not really know the advance flags or we should add a better support for OMC setting in OMEdit.

I prefer some limit based on  UNSOLVABLE,SOLVABLE, system size like

(number SOLVABLE in system)/( system size) > x%

which consider the chance for reduce system size by using tearing.

ptaeuber:
@vruge unsolvables directly get tearing variables so there is no compuation effort. however, using solvability criteria to assess the benefit of tearing of huge systems seems to be a good idea for further considerations.

@casella ok, I will introduce two flags, one for linear and one for nonlinear systems and will do some investigations with the ScalableTestSuite. For the runtime stuff I will open another PR.

casella:
@ptaeuber thanks!

@vruge I agree that a good default criterion should be more sophisticated than just comparing the size of the system with a threshold. This is a good research topic. However, a crude default is better than no default at all (i.e., infinity). If I am a naive user, I am trying to simulate a large system, and I don't know anything about advanced flags, OMC should at least avoid trying to tear a system with 100000 equations. 

The largest DistributionSystem examples in the ScalableTestSuite takes over one hour of back-end time, which I understand is mostly due to tearing, with simulation times below one minute - this makes no sense as a default behaviour, in particular if I am using OMEdit.

vruge:
@ptaeuber I mean unsolvables, solvables  for adjacency rows.

**e.g.** f( number of  each solvables adjacency rows  for each variable) vs. f( number of  each unsolvables adjacency rows  for each equations) 

**or** f( number of  each solvables adjacency rows  for each variable) vs. f( number of  each solvables adjacency rows  for each equations) 

where f(.) can be mean(.)

vruge:
@casella Good point, waiting without any rection from tool about 14,9 min is horrible....

sjoelund:
@vruge: Check some ScalableTestSuite [log](https://test.openmodelica.org/libraries/ScalableTestSuite_Experimental/files/ScalableTestSuite.Electrical.DistributionSystemDC.ScaledExperiments.DistributionSystemModelica_N_56_M_56.err). In the linked, log, 6766/7351 seconds is spent in tearing (initialization). With 1.845 TB / 1.975 TB of total memory allocated. If that one system is ignored (Linear torn systems: 1 {(3137,1.1%) 9518}), total performance should go up by a factor ~10x.

vruge:
@sjoelund 
Then, how about cpu-time-stop for tearing?

@ptaeuber 
It is possible:

e.g.
start tearing heuristic:
mark internal var <-> equation
if(not check_cpu_time()){
  (reset var,rest) -> (iterration_vars, resiuduum equations)
  return ;
 }
mark internal var <-> equation
if(not check_cpu_time()){
  (reset var,rest) -> (iterration_vars, resiuduum equations)
  return ;
 }

sjoelund:
@vruge checking CPU-time is rather expensive in a loop. Checking a maximum number of iterations is usually more efficient (and predictable for testing). But yes, there could be internal checks for progress in order to partially tear a system if you want to keep the progress you made so far.
Another alternative would be to see if the tearing module could be rewritten in order to reduce the memory/time it consumes.

ptaeuber:
So here is the current state of play:
- There are two compilation flags, one to set the maximum size for tearing of linear systems and one for tearing of nonlinear systems:
 - '+maxSizeLinearTearing=value'
 - '+maxSizeNonlinearTearing=value'
- The default value for nonlinear systems is 10000 and for linear systems 4000

- Then there are three new simulation flags to use sparse solvers for huge sparse systems
 - <-lss=value>
      value specifies the linear sparse solver method (default: umfpack)
 - <-lssMaxDensity=value>
      value specifies the maximum density for using a linear sparse solver (default: 0.2)
 - <-lssMinSize=value>
      value specifies the minimum system size for using a linear sparse solver (default: 4001)

- So for linear systems beyond sizes of 4000 tearing is disabled and sparse solvers activated. For nonlinear systems beyond 10000 only tearing is skipped. 

- The tearing flag and the sparse solver flags are not coupled, so it is still possible to handle torn systems with sparse solvers or nontorn (perhaps sparse) systems with nonsparse solvers.

- Moreover it is still possible to select a sparse solver with <-ls=value> for all systems without conditions (density, system size).

If a sparse solver is used because of size and density, e.g. the following info is printed:
`stdout            | info    | Using sparse solver for linear system 0,
|                 | |       | 
because density of 0.44 remains under threshold of 0.50 and size of 5 exceeds threshold of 2.
|                 | |       | 
The maximum density and the minimal system size for using sparse solvers can be specified
|                 | |       | using the runtime flags '<-lssMaxDensity=value>' and '<-lssMinSize=value>'.`
 
I think it is now pretty close to what is mentioned in [#3487](https://trac.openmodelica.org/OpenModelica/ticket/3487).

@casella @sjoelund What do you think? What about the defaults for this first realization?

sjoelund:
Were these thresholds acquired by benchmarking? I would have assumed the lss being used would depend on a function of size and density. A density of 0.2 I guess is considered very dense for matrices of size 4000 (but not perhaps 40). And I suspect matrices also of smaller sizes run fast using sparse solvers. @casella your thoughts?

casella:
On density vs. size I have no definite ideas. Maybe some literature search by Patrick on this topic could be useful. Unless there are some results in the literature we can use right away, I'd start experimenting and freeze the details later on once we've gathered some experience. BTW, I think the real advantages will show up for really large systems (100k or more equations) and with very low density (below 1%) - in those case, there is probably little question as to which is the optimal choice: no tearing, sparse solver.

I would also suggest to use KLU as linear default solver, instead of umfpack. For circuit models (which are most likely the origin of large linear algebraic models) KLU performs much better than umfpack.

ptaeuber:
KLU is default now.

OpenModelica-Hudson:
The job failed the coding style; for details, see [OpenModelica_TEST_PULL_REQUEST 2016-04-05_17-16-24](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2243/console). Remember to clean up the commit you did that breaks the coding style rather than piling a new commit on top of the old ones. This goes double for accidental addition of files that you need to remove.

OpenModelica-Hudson:
The job failed the coding style; for details, see [OpenModelica_TEST_PULL_REQUEST 2016-04-06_10-02-53](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2250/console). Remember to clean up the commit you did that breaks the coding style rather than piling a new commit on top of the old ones. This goes double for accidental addition of files that you need to remove.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-04-06_10-33-14](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2252/console).

PR #556, opened 2016-03-21
Add a removeVerySimpleEquations pre-opt module

sjoelund:
The reason for adding an additional early and faster remove simple
equations is that it speeds up the subsequent modules since they need
to traverse fewer equations and allocate less data. This gives upwards
of a 20% performance increase for medium-size models.


sjoelund:
@lochel @wibraun what do you think about these changes? It seems to improve performance a lot for models like ScalableTestSuite.Electrical.DistributionSystemDC.ScaledExperiments.DistributionSystemModelicaActiveLoads_N_28_M_28, which have removeSimpleEquations as one of the bottle-necks (the other being tearing, which should be fixed by #529 ).

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2016-03-21_16-29-01](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1991/).

wibraun:
Actually the module removeVerySimpleEquations coudl replace the removeSimpleEquations, but in removeSimpleEquations are a lot of implicit transformations done, which should moved out into a separated module. What I also started and want continue together with Mahder, because that involes a lot of array and record expansion. So but perhaps it's an idea to run this module in a different order, so start with removeVerySimpleEquations and and then run removeSimpleEquations.

sjoelund:
Yes, we need a really fast and specialized step for pre-processing the DAE. It seems that the BB removeSimpleEquations (which I used) still fails for some models, so perhaps an even simpler version module should be created. It should be enough to only detect scalar variables bound to a value or alias (linear traversal of eqs/vars), followed by a linear-time traversal of the remaining eqs/var, doing the recorded replacements.

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2016-03-21_17-23-30](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1994/).

sjoelund:
Some of the tests seem to produce better results for some reason. This [tearing example](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/1994/testReport/junit/(root)/simulation_modelica_tearing/Tearing3_omc_mos/) no longer needs tearing for initialization for example... But there are a lot of models changing, so I guess I should use an even simpler remove simple equations that does not remove anything it does not handle.

casella:
For very large models which are compiled and simulated once, anything with a complexity beyond linear simply does not pay off. I am strongly in favour of first running a bare-bone module with linear complexity. The more sophisticated modules can be run afterwards (and be more efficient, as the size of the system is reduced) or simply switched off when handling larger systems, possibly with some default threshold like in the case of tearing.

sjoelund:
Going beyond linear complexity is a necessity. Even sorting algorithms are `O(n*log(n))`. But `O(n*log(n))` is acceptable (at 4 billion elements, you scale 32 times worse than `O(n)`).

However, it is true that it is desirable to reduce the system quickly since if we can reduce `n` in linear time, `n*log(n)` will be smaller in subsequent steps.

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2016-03-23_11-55-20](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2016/).

casella:
@sjoelund I understand in many cases going beyond linear is necessary. But I guess in many cases it isn't. In particular, I would expect that the simcode phase scales proportionally to the number of nonzero entries in the incidence matrix of the system, as all the sorting, matching, tearing, ordering, simplifying etc. has already been carried out. Am I too optimistic in this belief?

sjoelund:
Yes, you would need to accept scaling the maximum of `O(n*log(n))` and the number of non-zero entries in the incidence matrix, I guess :)

casella:
Sound good. Where do I need to sign? ;)

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-03-23_16-30-13](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2027/console).

PR #564, opened 2016-03-23
Improve backend memory

sjoelund:


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-03-23_13-50-20](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2019/console).

sjoelund:
@casella this should improve performance of ScalableTestSuite quite a bit, even without the limits proposed in #529. Tearing still uses a lot of memory though.

casella:
It is interesting to see were the break-even size is now.
Furthermore, could we think of using a sparse solver on the residual equations when they are still very sparse?

sjoelund:
It would be interesting to see OM auto-select sparse/dense solvers depending on the density and size.

casella:
Absolutely! I think we can do some testing on the ScalableTestSuite and based on the results, select some reasonable default thresholds.

PR #572, opened 2016-03-24
Maximum number of event iterations

lochel:
This is about to fix [#3795](https://trac.openmodelica.org/OpenModelica/ticket/3795).


sjoelund:
I guess after you fix the commit names, etc. Just going to see if Hudson thinks you break something?

lochel:
I just implemented half of the stuff. However, I checked already if a decreased limit of 20 event iterations breaks any test of the testsuite and it does not.

lochel:
This PR contains following changes:
* decreased default limit of event iterations to `20`
* new flag `-mei=<max-event-iterations>`
* improved error message

@casella any comments?

casella:
@lochel That's excellent! I love these overnight fixes :)

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2016-03-24_10-20-42](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2041/).

casella:
Well,  in fact I do have some comments.

It becomes more and more important to implement [#3488](https://trac.openmodelica.org/OpenModelica/ticket/3488), otherwise, e.g., all the examples in the testsuite with more than 20 event iterations will always break. Of course we can set this flag in the settings of the entire coverage test, but in general it might be inappropriate to set -mei=1500 for other test cases.

Also, it is important to add some hints to the diagnostic messages as to how to fix the problem. E.g., add to the assert a phrase such as "You can increase the maximum number of event iterations by setting the flag -mei=<max-event-iterations>". Diagnostic messages will mostly be read by inexperienced users, rather than developers :)

@adeas31 It would be nice if this was somehow handled by the GUI in an assisted way, rather than the end user having to set some obscure flag in the Simulation tab or in the Options|Simulation tab

BTW (sorry for the off-topic comment) I still find it very confusing to understand which is which - eventually, both are named "Simulation flags", but I understand one collects the flags passed to the simulation runtime, the other the flags passed to OMC during compilation. An example with a couple of flags in the input box caption would help a lot. I'll open a ticket about that.

lochel:
> Also, it is important to add some hints to the diagnostic messages as to how to fix the problem. E.g., add to the assert a phrase such as "You can increase the maximum number of event iterations by setting the flag -mei=". Diagnostic messages will mostly be read by inexperienced users, rather than developers :)

I had already added such a hint:
```
$ ./EventIteration -mei=3
stdout            | info    | Maximum number of event iterations changed to 3
assert            | debug   | Simulation terminated due to too many, i.e. 3, event iterations. This could
either indicate an inconsistent system or an undersized limit of event iterations. The limit of event
iterations can be specified using the runtime flag '–mei=<value>'.
```

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-03-24_11-44-53](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2044/console).

casella:
Very good, thanks!

PR #574, opened 2016-03-24
fix modelicaStringToCStr for "'"

wibraun:


sjoelund:
I would rather make a general fix that handled escaping every possible name

wibraun:
How would you do that?

sjoelund:
A general approach like checking a regex if it is a valid C identifier and otherwise replace it with an escaped name, like we do for quoted identifiers:
```c
$P_omcQuot_72 = (2.0) * (data->localData[0]->timeValue); /* 'r' = 2*time */
```

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-05-17_21-32-59](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2687/console).

PR #589, opened 2016-03-28
Do not allocate tables for DSS using uode

sjoelund:
This commit disables allocating of large hash tables for dynamic state
selection when using uode (no DSS), and sizes the hash tables based on
the number of states if we do use DSS.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-03-28_18-32-42](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2100/console).

PR #592, opened 2016-03-28
Do not use Option<Name> in FGraph.openScope

sjoelund:
All calls to openScope were using SOME(id), which allocates a SOME-cell
for every single call to openScope. While easy to collect, it still
causes some extra work for the GC, so it has been refactored away.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-03-28_21-44-38](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2106/console).

PR #603, opened 2016-03-30
Fix SimulationRuntime/c for OSX

hkiel:


SimulationRuntime/c/util/rtclock.c
```diff
@@ -402,6 +402,23 @@ double rt_ext_tp_tock(rtclock_t* tick_tp) {
...
+}
+
+double rt_ext_tp_tock_realtime(rtclock_t* tick_tp) {
+  double d=0;
+  rtclock_t tock_tp = mach_absolute_time();
+  d = (double)(tock_tp - *tick_tp);
+  return d;
+}
+
+int64_t rt_ext_tp_sync_nanosec(rtclock_t* tick_tp, uint64_t nsec)
```
sjoelund: This function needs to sleep in some way in order to synchronize the clock to tick_tp + nsec number of nanoseconds. Else, I guess throw an error saying OSX cannot do this.

SimulationRuntime/c/util/rtclock.c
```diff
@@ -402,6 +402,23 @@ double rt_ext_tp_tock(rtclock_t* tick_tp) {
   return d - min_time;
 }
 
+void rt_ext_tp_tick_realtime(rtclock_t* tick_tp) {
+  *tick_tp = mach_absolute_time();
+}
+
+double rt_ext_tp_tock_realtime(rtclock_t* tick_tp) {
+  double d=0;
+  rtclock_t tock_tp = mach_absolute_time();
```
sjoelund: This would need to use `mach_timebase_info()` to return the delta in seconds between the two values.

hkiel:
Without the change I cannot simulate because the model does not compile (C compiler):
```
Undefined symbols for architecture x86_64:
  "_rt_ext_tp_sync_nanosec", referenced from:
      _fmtEmitStep in DistributionSystemDC.ScaledExperiments.DistributionSystemModelica_N_20_M_20.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

I do not want to implement the function, just get rid of the dynamic linker problem.

@sjoelund Can you fix the functions based on what you think is correct? (i.e. put the warrning, but do not fail when compiling the model)

sjoelund:
I cannot easily do that. No updated OSX system to test real-time synchronization.

```c
throwStreamPrint(NULL, "%s not implemented for OSX", __FUNCTION__);
```

hkiel:
@sjoelund Does that look better?

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-03-31_09-04-27](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2140/console).

PR #624, opened 2016-04-02
Do not generate function code for constants

sjoelund:
Constants are replaced by values and do not need to be duplicated in
functions. Currently, if evaluation of the constant fails, we get the
same constant that cannot be evaluated in the function prologue.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-04-02_05-59-35](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2176/console).

PR #628, opened 2016-04-03
Enable code generation in parallel

sjoelund:
The default will enable parallel code generation if you have more than
2 physical cores, and will use half of the physical cores for code
generation. The reason is that the templates allocate a lot of memory,
causing the GC to become slow if too many threads are used. On a 4-core
machine, 2 threads was 60~65% the speed of 1 thread and 3~4 threads around
75~105% the speed of 1 thread. It might need a new flag to tune the number
of threads used. -d=-parallelCodegen disables the new functionality.
NOTE: Parallel code generation is only used for the C code generator.

### DIFF
--- a/Compiler/SimCode/SimCodeMain.mo
+++ b/Compiler/SimCode/SimCodeMain.mo
@@ -75,6 +75,7 @@ import Debug;
 import Error;
 import Flags;
 import FMI;
+import GC;
 import HpcOmSimCodeMain;
 import HpcOmTaskGraph;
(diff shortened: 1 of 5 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-04-03_13-57-17](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2195/console).

PR #632, opened 2016-04-03
Handle listReverse of 1 element as a special case

sjoelund:


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2016-04-03_16-53-10](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2199/).

sjoelund:
I guess some Dangerous.listReverse or similar gets a listReverse'd list and assumes it's all fresh cells that are safe to destroy causing the error.

hkiel:
append_reverse() uses listReverse() and I always assumed that it is safe to do listReverseInPlace() after using append_reverse() starting with an empty list.
I'm not sure which one gives better results:
using append_reverse()+listReverseInplace() and listReverse() "reversing" the single-element-list
or using append_reverse()+listReverse() and listReverse() NOT "reversing" the single-element-list...

hkiel:
@sjoelund I guess I shall revert to
append_reverse()+listReverse()

sjoelund:
Maybe. I don't know which is better. Possibly making append_reverse not do the listReverse internally as well. I can play around with it since I see a potential problem. When I have time. I just wanted to know how much one would gain by making listReverse of 1 element not allocate memory actually saves.

hkiel:
I counted the single-element lists in one example and it counted ~160.000 (simulation/libraries/msl32/log-Modelica.Blocks.Examples.ShowLogicalSources.mos)
So, maybe it is worth trying it.

I will commit a patch so that append_reverse is no longer used with listReverseInPlace (hopefully found all ocurrences)

sjoelund:
I will run Hudson PUSH=false to see if 30ff47b helps. If it does, I guess the other functions do not need to be changed although it would be good if the function was somehow renamed to signify that it will always copy the elements.

hkiel:
My bet is that it will work, because I could fix the failing tests by using listReverse() after append_reverse

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-04-04_13-24-56](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2223/console).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-04-04_14-06-15](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2224/console).

PR #636, opened 2016-04-04
Split up knownVars

lochel:
`knownVars` should only contain those variables, that are constant for the entire simulation.
The newly introduced `globalVars` can be used for variables that are somehow known but not constant for the entire simulation.

### DIFF
--- a/Compiler/BackEnd/BackendDAE.mo
+++ b/Compiler/BackEnd/BackendDAE.mo
@@ -106,6 +106,7 @@ public
 uniontype Shared "Data shared for all equation-systems"
   record SHARED
     Variables knownVars                     "Known variables, i.e. constants and parameters";
+    Variables globalVars                    "Global variables, i.e. time, state, or input dependent";
     Variables externalObjects               "External object variables";
     Variables aliasVars                     "Data originating from removed simple equations needed to build
                                              variables' lookup table (in C output).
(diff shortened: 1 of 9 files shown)


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2016-04-04_11-39-30](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2219/).

lochel:
It's just the optimization that has issues. @vruge Can you fix that?

vruge:
no

vruge:
I guess it is line 738 in DynamicOptimization
BackendDAE.SHARED(functionTree = funcs, knownVars = knownVars) := shared;
missing globalVars??

vruge:
additional  RemoveSimpleEquations.removeSimpleEquationsShared need globalVars

vruge:
and SymbolicJacobian.generateSymbolicJacobian

lochel:
@vruge It would be good if you could at least fix the optimization, since I don’t get your code right.

vruge:
why do you need an additional explanation with "It would be good if"?
Do you think the  "no" was gratuitous?

lochel:
Unfortunately, that will lead to nothing.

PR #639, opened 2016-04-04
Use append_reverse with listReverse only

hkiel:
NEVER use listReverseInPlace, as listReverse (that is used by append_reverse) may not really reverse a single element list

That should help #632 to work (it does on my machine for the failed test)


hkiel:
Martin found a better solution. (fix append_reverse to not use listAppend() directly, but the algorithmic equivalent (without optimizations))

PR #646, opened 2016-04-05
Introduce runtime flag -overrideSettings

lochel:
see [ticket:3810](https://trac.openmodelica.org/OpenModelica/ticket/3810)


lochel:
Should it be possible to override a given setting multiple times, e.g., `-overrideSettings=startTime=2.0,startTime=3.0`?

How to deal with such cases?
a) use the last value
b) throw a warning/error

PR #651, opened 2016-04-05
avoid temporary intermediate lists

hkiel:
instead use:
List.fold(List.map()) -> List.applyAndFold()
List.map(List.map()) -> List.mapMap()
listReverse(List.mapFlat()) -> List.mapFlatReverse()

### DIFF
--- a/Compiler/BackEnd/BackendDAETransform.mo
+++ b/Compiler/BackEnd/BackendDAETransform.mo
@@ -253,7 +253,7 @@ algorithm
       var_lst = List.map(var_varindx_lst, Util.tuple21);
       crlst = List.map(var_lst,BackendVariable.varCref);
        // its only an array equation if all the solved variables belong to an array. Otherwise we have to handle it as a non-linear system
-      b1 =  List.fold(List.map(crlst,ComponentReference.isArrayElement),boolAnd,true);
+      b1 =  List.applyAndFold(crlst,boolAnd,ComponentReference.isArrayElement,true);
       if not b1 then
         expLst = List.map(crlst, Expression.crefExp);
         true = List.exist1(inEqnLst,crefsAreArray,expLst);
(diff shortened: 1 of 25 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-04-05_16-28-13](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2242/console).

sjoelund:
@hkiel if you want some guidance for which parts of the code consume most memory, see: https://dev.openmodelica.org/~marsj/tmp/20160405-memory.txt (line numbers should match 6c0fc0ddad05d79ffd3ce106889acbf98c4f0619 and has merged allocations in utility packages into their caller)

casella:
@hkiel for elapsed times, instead, you can look at the log files linked from https://test.openmodelica.org/libraries/ScalableTestSuite_Experimental/BuildModelRecursive.html

hkiel:
thanks for the input.
When I spot an inefficiency, I tend to fix it everywhere independent on how much impact it will have.
I will have a look at the numbers. I already checked the ScalableTestSuite (which I cannot load fully on my OSX) and did some XCode:instrument to find out recursion depths and running times of a few models.

casella:
You mean you cannot compile it fully, or just load it? The DistributionSystemModelicaIndividual_N_80_M_80 has about 20k components and 20k connect statements, but it shouldn't be a problem to just load it. Is it?

hkiel:
@casella I just noticed I can loadFile() it right now. That failed few weeks ago. Maybe memory usage improved now...

casella:
Very good! Maybe you also have stack overflow problems with OSX, in Linux I had to increase the stack size to the maximum allowed value (i.e. 64 MB) to run our network models. I am not sure what is the default for the Mac, maybe it is really too low

sjoelund:
OSX has more stack overflow issues than Linux. Not because the stack size was small but because the stack frames were larger.

hkiel:
@sjoelund 
From what I see in memory.txt file ConnectUtil.mo line1171 uses List.replaceAt() and this is very expensive.
Can you at some point change from list of nodes to array of nodes and then apply all replaceAt (by cheaper arrayUpdate) and if needed convert back to list or stay with array operations?

I do not get where adding ends and where replacing starts...

perost:
I am working on fixing the connection handling, which will require massive changes to ConnectUtil and other things. So please ignore ConnectUtil for now.

PR #661, opened 2016-04-06
Memory leak fixes for sparse solvers.

crupp2:
Fixes for memory leaks in the sparse linear solvers, both umfpack and klu. See ticket:3548.

In umfpack, from everything I've looked at it looks like you must free the Numeric object whenever the matrix changes, which for this application is at every linear solve. I've added this and also changed the solve to wsolve, which uses a workspace to save on some malloc/free calls.

The klu solver also leaked due to a similar issue. However, it has a refactor call that can reuse memory and is a lot faster than factor, but it uses the same pivoting as the previous factor call, which may be unstable as the matrix changes during the simulation. So, I added some logic to check that refactor does ok (via rgrowth) and if not then do a full factor. The questionable thing is what the rgrowth tolerance should be. The guidance from documentation is "If rgrowth is very small, an inaccurate factorization may have been performed." I did some testing and 1e-3 seems fine and probably safe, but it maybe could go lower. Alternatively, condest or rcond could be used.

### DIFF
--- a/SimulationRuntime/c/simulation/solver/linearSolverKlu.c
+++ b/SimulationRuntime/c/simulation/solver/linearSolverKlu.c
@@ -258,7 +258,20 @@ solveKlu(DATA *data, threadData_t *threadData, int sysNumber)
 
   /* compute the LU factorization of A */
   if (0 == solverData->common.status){
-    solverData->numeric = klu_factor(solverData->Ap, solverData->Ai, solverData->Ax, solverData->symbolic, &solverData->common);
+    if(solverData->numeric){
+      /* Just refactor using the same pivots, but check that the refactor is still accurate */
+      klu_refactor(solverData->Ap, solverData->Ai, solverData->Ax, solverData->symbolic, solverData->numeric, &solverData->common);
+      klu_rgrowth(solverData->Ap, solverData->Ai, solverData->Ax, solverData->symbolic, solverData->numeric, &solverData->common);
+      infoStreamPrint(LOG_LS_V, 0, "Klu rgrowth after refactor: %f", solverData->common.rgrowth);
+      /* If rgrowth is small then do a whole factorization with new pivots (What should this tolerance be?) */
+      if (solverData->common.rgrowth < 1e-3){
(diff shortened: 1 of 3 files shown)


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2016-04-11_17-10-42](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2351/).

casella:
Why is the result of the pull request testing only available after 5 days?

casella:
See also [ticket:3646](https://trac.openmodelica.org/OpenModelica/ticket/3646)

perost:
>Why is the result of the pull request testing only available after 5 days?

Pull requests are not tested automatically, the Hudson job is started manually by a developer. So the tests results are available now simply because @wibraun ran the Hudson job two hours ago.

crupp2:
It looks like the test results are due to something other than the code changes. The changes I made wouldn't have affected path locations or model naming.

I do expect small differences in the results, however. Although everything I saw was in the last few decimal places.

wibraun:
@crupp2 Your code change are good and correct. I try to push also some more tests for the sparse solvers and fail on our rigoros testsuite policy ;).
In the first run I missed to add a file and in the second I added it but hudson doesn't like that file at that place. However, I'm sure I can convince Hudson to take this changes, but I do now a break.

crupp2:
@wibraun Ok, sounds good. Thanks for being rigorous!

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-04-11_23-14-17](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2361/console).

wibraun:
@casella I don't think this fixes [ticket:3646](https://trac.openmodelica.org/OpenModelica/ticket/3646), since this was at least for KLU a performance enhancement, although a good one in my short tests I measured improvements by factor ~2.

PR #665, opened 2016-04-07
Make error messages less verbose for performance

sjoelund:
Do not create a prefixed path to string for every single variable that
is instantiated. Possibility: use a debugging flag to turn on more
diagnostics. This *should* be OK since we pass along SourceInfo to most
of the functions in the compiler, which is the main task of
Error.updateCurrentComponent. Also getting the full instance name
should be a minor detail.

### DIFF
--- a/Compiler/FrontEnd/InstVar.mo
+++ b/Compiler/FrontEnd/InstVar.mo
@@ -571,8 +571,8 @@ protected
   DAE.ElementSource source;
 algorithm
   try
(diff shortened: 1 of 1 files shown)


hkiel:
Why is this closed?

sjoelund:
I'll do a better implementation after talking to Adrian, see https://trac.openmodelica.org/OpenModelica/ticket/3824

PR #666, opened 2016-04-07
Make error messages consume less memory

sjoelund:
This implements ticket:3824 by making updateCurrentComponent store the
Prefix instead of always making a String out of it. This should result
in a big performance gain without changing the behaviour of OMC.

### DIFF
--- a/Compiler/FrontEnd/Inst.mo
+++ b/Compiler/FrontEnd/Inst.mo
@@ -3226,7 +3226,7 @@ algorithm
     (outCache, outEnv, outIH, outStore, DAE.DAE(outDae), outSets, outState, outVars, outGraph, outFieldDomOpt) :=
       instElement(outCache, outEnv, outIH, outStore, inMod, inPrefix, outState, elt, inInstDims,
         inImplicit, inCallingScope, outGraph, inSets);
-    Error.updateCurrentComponent("", Absyn.dummyInfo);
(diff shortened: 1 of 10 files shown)


OpenModelica-Hudson:
The job failed to compile; for details, see [OpenModelica_TEST_PULL_REQUEST 2016-04-07_11-17-26](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2274/console).

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2016-04-07_11-25-14](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2275/).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-04-07_11-42-31](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2276/console).

sjoelund:
9% performance improvement while giving the same test suite output as before.

hkiel:
I wonder how a model without errors will gain from this patch.
At least the test suite runs faster.

sjoelund:
Quite simple: `PrefixUtil.prefixPath(Absyn.IDENT(inName), inPrefix))` is deferred until there is an error that does not use source information (so Error.addMessage instead of Error.addSourceMessage). There are very few of those instances in the code. Perhaps we could even have removed all of this logic (only 5 tests in the entire test suite actually outputs the generated information; barely worth the trouble).

PR #669, opened 2016-04-07
Fix bottleneck for create model info

sjoelund:
This is related to ticket:3813. Uses an array of SimVar instead of
pushing everything into a large uniontype record for every change
to a simvar. The array is indexed by an enumeration, which maintains
some knowledge of which field a particular index belongs to. The
enumeration has elements ordered in FMI order since it is important
these indexes appear in a certain order.

### DIFF
--- a/Compiler/SimCode/SimCodeUtil.mo
+++ b/Compiler/SimCode/SimCodeUtil.mo
@@ -6273,6 +6273,45 @@ algorithm
   end matchcontinue;
 end getStateInfo;
 
+protected type SimVarsIndex = enumeration(
+  // In special order for fmi: real => intger => boolean => string => external
+  state,
+  derivative,
+  alg,
+  discreteAlg,
+  param,
+  alias,
+
+  intAlg,
+  intParam,
+  intAlias,
+
+  boolAlg,
+  boolParam,
+  boolAlias,
+
+  stringAlg,
+  stringParam,
+  stringAlias,
+
+  extObj,
+
+  inputs,
+  outputs,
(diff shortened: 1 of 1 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-04-07_16-17-24](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2283/console).

hkiel:
I you compare times in the TEST_PULL_REQUEST logs from previous (2282) to this version (2283), then you'll see that things got slower. Any idea why?

wibraun:
wow, this looks quite pretty.

sjoelund:
I'm not sure why it did not get faster. Maybe I sorted one array too many in some place. It should be (a lot) faster, and models from @casella did consume less memory as well. But it only really scales better for larger models. And performance of the test suite between 2282 and 2283 actually did get faster (within the margin of error). I compare the total wall-time taken for the runtests.pl command. I guess you looked at the total aggregate time for all (serial tests), which increased (but those times are summed up from integer seconds for tests, so they are inaccurate).

sjoelund:
@wibraun Yes, we should use enumerations more often. It's a bit bad we can't index MetaModelica arrays with enumerations directly though (and Modelica-style arrays do not give good enough performance).

casella:
The impact on the RETE_D model was dramatic: from 147.3 s down to 32.3. I wonder if there is some more room for improvement

PR #693, opened 2016-04-11
PDEModelica in AbsynDump

jansilar:
field and indomain keywords added to AbsynDumpTpl templates. Fixing ticket: 3755

### DIFF
--- a/Compiler/Template/AbsynDumpTV.mo
+++ b/Compiler/Template/AbsynDumpTV.mo
@@ -461,10 +461,16 @@ package Absyn
       Boolean streamPrefix;
       Parallelism parallelism;
       Variability variability;
+      IsField isField;
       Direction direction;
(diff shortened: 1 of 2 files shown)


sjoelund:
Starting the testsuite now. Do you have any tests to add to the test suite so this does not break in the future?

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-04-13_10-54-16](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2388/console).

jansilar:
Thanks for integrating. I think we don't need any test particularly for this. It was not an unintentionally introduced bug. The problem was here from the beginning of the extension. I just forgot to implement support for new keywords in the AbsynDump.

PR #716, opened 2016-04-13
Fix name of OPC UA aliases

sjoelund:
Previously, we put the name of the actual variable instead of the alias.
Also, we added aliases for parameters even though parameters are not
added in the OPC UA interface.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-04-13_10-20-37](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2387/console).

PR #717, opened 2016-04-13
Clean up DAEUtil.splitElements.

perost:
- Clean up DAEUtil.splitElements.
- Move DAEDump.dumpDAEList -> DAEUtil.splitComponent

### DIFF
--- a/Compiler/FrontEnd/DAEUtil.mo
+++ b/Compiler/FrontEnd/DAEUtil.mo
@@ -4950,23 +4950,121 @@ algorithm
 end joinDaeLst;
 
 public function splitElements
-"@author: adrpo
-  This function will split DAE elements into:
-   variables, initial equations, initial algorithms,
-   equations, algorithms, constraints and external objects"
-  input list<DAE.Element> inElements;
-  output list<DAE.Element> v;
-  output list<DAE.Element> ie;
-  output list<DAE.Element> ia;
-  output list<DAE.Element> e;
-  output list<DAE.Element> a;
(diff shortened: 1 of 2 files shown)


sjoelund:
There are also functions that call splitElements without using some of the outputs (so isPresent could be used to conditionally not create some of the lists). In particular, there is one call that only filters algorithms (which perhaps should be a separate function or just calling a filter directly... list(e for e guard isNonInitialAlgorithm(e) in elts))?

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-04-13_11-08-48](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2389/console).

Compiler/FrontEnd/DAEUtil.mo
```diff
@@ -4950,23 +4950,121 @@ algorithm
...
+
+  variables := listReverse(variables);
+  initialEquations := listReverse(initialEquations);
+  initialAlgorithms := listReverse(initialAlgorithms);
+  equations := listReverse(equations);
+  algorithms := listReverse(algorithms);
+  classAttributes := listReverse(classAttributes);
+  constraints := listReverse(constraints);
+  externalObjects := listReverse(externalObjects);
+  stateMachineComps := listReverse(stateMachineComps);
```
hkiel: you built the lists from scratch by prepending elements, so you can safely use MetaModelica.Dangerous.listReverseInPlace() here (and in the lines above)

PR #724, opened 2016-04-14
update the output of the numerical linearization

jhag:
I got conflicts when I enabled my module `wrapFunctionCalls `with the test models of `/openmodelica/linearization/`...
The reason is that the numerical linearization writes the cse-variables with dollar signs in `.mo-files`, e.g. `$cse1`. That is not allowed in the Modelica language.
@crupp2  So it is okay if I put the cse-variables and others in quotation marks ` '$cse1' ` ?

### DIFF
--- a/Compiler/Template/CodegenC.tpl
+++ b/Compiler/Template/CodegenC.tpl
@@ -4122,11 +4122,11 @@ template functionlinearmodel(ModelInfo modelInfo, String modelNamePrefix) "templ
     let vectorU = genVector("u", varInfo.numInVars, 1)
     let vectorY = genVector("y", varInfo.numOutVars, 2)
     let vectorZ = genVector("z", varInfo.numAlgVars, 2)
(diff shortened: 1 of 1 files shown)


sjoelund:
I would say making them quoted identifiers is OK, but you then must check that the variable does not already exist. Using a $ prefix is easier since you know this is an invalid Modelica identifer and no user will have introduced such a name.

lochel:
The quoted identifiers will only show up in the generated linear model. So I guess that it isn’t possible to get name conflicts.

crupp2:
I am fine with that. Thanks for including me. :)

In fact, if all the variables are now going to be in quotes then is it also possible to use the correct variable name rather than the pseudo-escaped junk (i.e., 'P' for '.', 'lB' for '[', etc.)? Currently it is a pain to parse the linearized variable names and match them to the original model because the mapping is not unique. For example, 'x_myParseVar' could be either 'x_myParseVar' or 'x_my.arseVar'. The only way around this is to cross check with the variable list, but it is much easier if that is not needed.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-04-14_23-05-20](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2421/console).

PR #749, opened 2016-04-18
Use referenceEq more for InstExtends/Absyn

sjoelund:
Still does not find every possible structure to use referenceEq on, but
does find a much larger number than before (before, CREFs, function
calls, modifiers, etc. all caused referenceEq to fail).


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-04-18_17-34-16](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2466/console).

PR #754, opened 2016-04-20
use list comprehension instead of for-loop+listReverse

hkiel:
Please review before merging!

Resulting C code is not better than before, but may run with same speed when C compiler optimizes, at the cost of slower code compilation.
Maybe the C Template should produce more optimal code for list-comprehensions:
```
/* list-comprehension: */
    while(1) {
      tmp2 = 1;
      while (!listEmpty(lst_loopVar)) {
        _lst = MMC_CAR(lst_loopVar);
        lst_loopVar = MMC_CDR(lst_loopVar);
        if(1) { /* found non-guarded */
          tmp2--;
          break;
        }
      }
      if (tmp2 == 0) {
        _$tmpVar0 = listLength(_lst);
        _$tmpVar1 = (modelica_integer)_$tmpVar1 + (modelica_integer)_$tmpVar0;
      } else if (tmp2 == 1) {
        break;
      } else {
        MMC_THROW_INTERNAL();
      }
    }
```

```
/* for loop:  (requires listReverse) */
    for (tmpMeta[0] = _inList1; !listEmpty(tmpMeta[0]); tmpMeta[0]=MMC_CDR(tmpMeta[0]))
    {
      _e = MMC_CAR(tmpMeta[0]);
      tmpMeta[1] = mmc_mk_cons(_e, _outList);
      _outList = tmpMeta[1];
    }
```
At least listReverseInPlace is not performed...

Next step would be to remove map#Fold() functions and use closure instead for the constant parameters.

### DIFF
--- a/Compiler/Util/List.mo
+++ b/Compiler/Util/List.mo
@@ -3660,7 +3660,7 @@ public function mapFold<TI, TO, FT>
   input list<TI> inList;
   input FuncType inFunc;
   input FT inArg;
-  output list<TO> outList = {};
+  output list<TO> outList;
   output FT outArg = inArg;
 
   partial function FuncType
@@ -3669,14 +3669,15 @@ public function mapFold<TI, TO, FT>
     output TO outResult;
     output FT outArg;
   end FuncType;
-protected
-  TO res;
 algorithm
-  for e in inList loop
-    (res, outArg) := inFunc(e, outArg);
-    outList := res :: outList;
-  end for;
-  outList := listReverseInPlace(outList);
+  outList := list(match e
+    local
+      TO res;
+      case _
+        algorithm
+        (res, outArg) := inFunc(e, outArg);
+        then res;
+    end match for e in inList);
 end mapFold;
 
 public function mapFold2<TI, TO, FT1, FT2>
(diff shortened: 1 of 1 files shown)


sjoelund:
A note about using closures: it costs a memory allocation to use a closure, so they are slower in particular when the list is empty (which is quite frequent). The same performance as list comprehension can be gotten by using DoubleEndedList.mo (it can also build lists in the "wrong" order).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-04-20_11-06-27](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2483/console).

hkiel:
@sjoelund Did you have a look at the generated C code?
Are we going to optimize that?

sjoelund:
We should not optimize in the current generation of the code generator. It will run fast enough in most cases.

hkiel:
What would be the next step?
Optimize current generated code or write a new code generator (maybe using different language)?

Also, is my patch to use "list-comprehension in conjunction with match" better than "for-loop plus listReverseInPlace"??? I do not see any speedup, but that may have been eaten up by the C-compiler taking longer to process unoptimized code...

PR #761, opened 2016-04-20
Construct CREF after checking HT is not empty

sjoelund:
Constructing a CREF from Prefix is very expensive. It should probably
not be done at all (HashSet from ComponentPrefix to X instead of always
creating a CREF from Prefix first).


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-04-20_21-35-05](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2497/console).

PR #762, opened 2016-04-20
Make special case for BackendVariable.getVarSingle

sjoelund:
BackendVariable.getVar returns a list of often size 1, and is expected
to return a list of exactly size 1. This checks that the size is 1
directly in the call to getVarSingle. Also uses isPresent to avoid
constructing the indexes list if it is not desired.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-04-20_22-31-03](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2498/console).

PR #773, opened 2016-04-21
Add LEAF node for AVL trees (half memory usage)

sjoelund:


OpenModelica-Hudson:
The job failed to compile; for details, see [OpenModelica_TEST_PULL_REQUEST 2016-04-21_14-33-21](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2513/console).

Compiler/Util/BaseAvlSet.mo
```diff
@@ -134,6 +156,7 @@ algorithm
   key := match inTree
     case EMPTY() algorithm return; then fail();
     case NODE() then inTree.key;
+    case LEAF() then inTree.key;
```
hkiel: better do `else inTree.key;` instead of two cases NODE() and LEAF()?

sjoelund: That would be a compile-time error (because we do not know if it is NODE or LEAF

hkiel: Then in (new) line 167 add `else false;` to account for the new record NODE(), otherwise the match fails.

sjoelund: True. It should not fail...

OpenModelica-Hudson:
The job failed to compile; for details, see [OpenModelica_TEST_PULL_REQUEST 2016-04-21_17-10-54](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2514/console).

hkiel:
applying the patch locally works fine... (did not try bootstrapping)

hkiel:
Any try to make this work?
My calculations assume less than half the memory usage (balanced binary tree has ~50% leaves and 50% inner nodes, so for half of the total nodes you use 2 instead of 5 words, makes a reduction of 30% in memory)

sjoelund:
Working on MODRIO deliverables. I will try to make it work. Note that the number of words used by a uniontype is 2+n where n is the number of fields. So 3 vs. 6 for sets, 4 vs. 7 for trees. However, GC allocates in chunks of certain sizes (8/16 bytes; I guess 2 words on 32/64-bit platforms, not sure) so it might be actually using 4 vs. 6 for sets and 4 vs. 8 for trees.

For example, GC reports 104.6 GB memory allocations while we only requested a total of 83695461544  bytes (77.9 GB) of allocations. I suspect a lot of the memory allocations are lists, which are 3 words (with an overhead ~33%).

sjoelund:
The Tree.get did not have a case for LEAF

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2016-04-26_11-05-39](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2538/).

sjoelund:
And a problem with the rotate functions. Also, when rotating, NODE was always created even when it was rotated into a LEAF.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-04-26_12-38-33](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2540/console).

PR #777, opened 2016-04-22
Use templates to write to file directly

sjoelund:
Used misc. templates in other templates instead of importing the string
conversion function in the templates. The output changes slightly since
we can no longer trim strings, etc.

Also introduced the possibility to use File and Tpl.Text objects in the
same package (used for JSON serialization, which is mostly directly to
file but also needs to printExpStr).


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2016-04-22_21-43-08](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2522/).

PR #779, opened 2016-04-22
Fix initialization of dynamic state selection

lochel:
There are cases were the dynamic state selection cannot get initialized due to odd dependencies to the simulation system. Therefore, it could be a good idea to move the initialization of the dynamic state selection behind the initial evaluation of the simulation system.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-04-22_20-39-59](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2520/console).

PR #792, opened 2016-04-27
wrapFunctionCalls is enabled postOpt

jhag:
This module introduces variables for each function call and substitutes all these calls with the newly introduced variables. Some special cases are skipped and variables are not introduced.

### DIFF
--- a/Compiler/FrontEnd/Expression.mo
+++ b/Compiler/FrontEnd/Expression.mo
@@ -9771,7 +9771,7 @@ public function isExpIfExp
   output Boolean res;
 algorithm
   res := match(e)
-    case(DAE.IFEXP()) then true;
+    case DAE.IFEXP() then true;
     else false;
   end match;
 end isExpIfExp;
--- a/Compiler/BackEnd/CommonSubExpression.mo
+++ b/Compiler/BackEnd/CommonSubExpression.mo
@@ -64,11 +64,12 @@ protected import ResolveLoops;
 protected import SynchronousFeatures;
 protected import Types;
 
+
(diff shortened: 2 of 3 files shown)


sjoelund:
You should squash/fixup your commits I guess. Or do you only want to see if your changes pass the tests?

lochel:
Yes, this should get squashed to a single commit before pushing it. However, we can first run Hudson to check the testsuite.

Compiler/FrontEnd/Expression.mo
```diff
@@ -9766,12 +9766,12 @@ algorithm
 end isExpCrefOrIfExp;
 
 public function isExpIfExp
-"Returns true if expression is an if expression"
+"Returns true if expression is an if or a smooth expression"
```
lochel: Is that correct?

jhag: Sorry, this line is wrong! I remove it when I fix it tomorrow!

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-04-28_10-46-58](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2567/console).

jhag:
We push it under reserve because we want to see the effects of the coverage test.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-04-28_12-03-54](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2570/console).

PR #807, opened 2016-05-03
Don't copy unit information from state to state derivative

lochel:
See [ticket:3668](https://trac.openmodelica.org/OpenModelica/ticket/3886).


lochel:
It would be much better to derive the proper unit information. However, this should be better than a wrong unit information anyway.

sjoelund:
Note: it is very slow to do:
```mo
a.x := 1
a.y := 2
a.z := 3
```
It costs n memory allocations of size `m*sizeof(void*)*(fields+2)`

lochel:
I see! However, that's something for another PR.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-05-03_11-50-30](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2607/console).

hkiel:
There is already a trac item for this: [3814](https://trac.openmodelica.org/OpenModelica/ticket/3814)

PR #824, opened 2016-05-10
ticket:3488

adeas31:
Read the __OpenModelica_commandLineOptions annotation and apply it.

### DIFF
--- a/Compiler/Script/CevalScriptBackend.mo
+++ b/Compiler/Script/CevalScriptBackend.mo
@@ -2689,21 +2689,33 @@ algorithm
       list<String> libs;
       String file_dir, fileNamePrefix;
       Absyn.Program p;
+      Flags.Flags flags;
       String commandLineOptions;
       list<String> args;
+      Boolean haveAnnotation;
 
     case (cache,env,_,st as GlobalScript.SYMBOLTABLE(),fileNamePrefix,_,_)
-      equation
+      algorithm
         // read the __OpenModelica_commandLineOptions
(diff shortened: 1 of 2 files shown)


adeas31:
The only problem with the above code is that the flags sets via `__OpenModelica_commandLineOptions` are saved permanently. So we need something like,

- `getCommandLineOptions` to get the existing flags
- Apply the new flags retrieved from `__OpenModelica_commandLineOptions`
- Reset to the old flags.

@perost can you write `getCommandLineOptions`?

sjoelund:
What's wrong with `Flags.saveFlags` followed by `Flags.loadFlags`? You just need to consider all execution paths (probably with a try-block) to avoid accidentally keeping the new flags.

adeas31:
Yeah I expected that to work and tried it the other way around but it doesn't work,

```
flags = Flags.loadFlags();
read/apply new flags and do translations
Flags.saveFlags(flags);
```

What do you mean by execution paths?

sjoelund:
Execution paths = control flow
```mo
try
  Flags.saveFlags();
  // apply new flags
  // do the stuff you wanted to do
  Flags.loadFlags(); // Resets to the original flags
else
  Flags.loadFlags(); // Handle a failed "stuff you wanted to do"; the reset is now always performed
  fail();
end try;
```

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-05-12_12-55-10](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2659/console).

PR #829, opened 2016-05-12
optimize addEquations for less Array.expand() calls

hkiel:
clean up addEquation


hkiel:
This should help if list is longer than 0.4xArraySize.
@sjoelund  Does it make sense to include this patch?

PR #853, opened 2016-05-17
don't add noEvent twice

vruge:
we already add noEvent inside inline


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-05-18_00-08-40](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2693/console).

PR #862, opened 2016-05-19
- dump a graphic blt-matrix in a html file via +d=bltmatrixdump

vwaurich:


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-05-19_13-37-36](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2716/console).

hkiel:
The only problem is that it does not show nicely (or even at all) on firefox
iexplorer works, but is not available on Linux

vwaurich:
On Firefox, the canvas size is quite limited. For smaller systems it works fine on my machine with Firefox. IE worked even for the biggest  matrix of Engine V6.
Any suggestions how this problem could be adresssed?

casella:
For larger systems, you may consider dumping a .png file, which is also a W3C format. 
The GIMP can handle .png images of up to  262144 by 262144 pixels. That won't be enough for my network models with half a million equations, but I guess those are too big for a graphical representation anyway.

sjoelund:
But PNG is a binary format. I don't think he wants to dump that directly from OMC :) Else, a dot-file or something and converting it to PNG/SVG/etc

vwaurich:
I liked html because it allows some interaction like displaying additional information when hovering over things or jumping to corresponding debug-output. images are nice but I would like to use this for debugging purposes as well (we'll see how practical this will be)

casella:
I see. What about using multiple canvases?

PR #866, opened 2016-05-20
Add flag to change the tolerances for newton solver

ptaeuber:
<-newtonFTol=value> or <-newtonFTol value>
  [double (default 1e-24)] tolerance for accepting accuracy in Newton solver
<-newtonXTol=value> or <-newtonXTol value>
  [double (default 1e-24)] tolerance for updating solution vector in Newton solver


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-05-20_11-05-18](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2722/console).

casella:
Thanks Patrick!
Since the machine eps in double precision is around 1e-16, I am a bit surprised by this 1e-24. Can you please explain quickly what these two numbers mean?
Anyway, we tried a couple of simulations with 1e-22 instead of 1e-24, and already got a 2X speed improvement...

wibraun:
We use there he 2. norm, but without the square operation, so 1e-24 is actually 1e-12 with ordinary 2-norm.

casella:
OK. Using the 2-norm would probably be clearer, maybe we should change this now before we run into backward compatibility issues.

casella:
Can you please briefly comment on how these thresholds are used?

lochel:
It would be good to use 2-norm based thresholds instead of the squared ones. At least for user interaction.

wibraun:
The basic check is the following:
`(((errorf <ftol) || (error_f_scaled < ftol)) && ((delta_x_scaled < xtol) || (delta_x < xtol)))`
where 
`error_f = norm(residuals)`
`error_f_scaled = norm(resScaling * residuals)`
and where 
`ResScaling = matrixNorm(Jac)`.

`delta_x = norm(x_{j} - x_{j-1})`
`delta_x_scaled= norm(xScaling * residuals)`
and where 
`xScaling = max(nominal(x_{i}, abs(x0{i})`

PR #872, opened 2016-05-23
Replace old new instantiation with new new instantiation.

perost:


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-05-23_10-59-41](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2733/console).

casella:
Did we already have a new instantiation, although old? I thought all new stuff was just in private branches.
Should we expect significant performance changes after this, in terms of coverage and speed?

adrpo:
@casella: Yes, we did had some previous implementation but it was not completed (mostly because some things were not at all clear in the specification and the implementation could not be fixed). Now that the specification is a bit more clear (see https://trac.modelica.org/Modelica/ticket/1829) we can restart the work based on a new prototype from @perost but it will take a while to have something that will be usable.

perost:
@casella: And to be clear, the old code wasn't actually used by anything, so it might as well have been in a private branch. But this was back in the days of SVN, where private branches weren't really a thing. The new code used to be in my own private branch, but is now replacing the old code in master to allow more people to work on it. Don't read too much into it, we're just moving some code around.

PR #917, opened 2016-05-27
Performance fix of instantiation of syst in loop

ptaeuber:
@hkiel Is this way ok?


hkiel:
looks, but I do not understand line 3681...

ptaeuber:
@hkiel It is to set up a new variables structure because the function removeVar does not work properly since "numberOfElements" is not reduced. The fix breaks a lot of models so this bad workaround is used for now. The module is not activated anyway and its only purpose is to test the new global vars structure to avoid that somebody breaks it.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-05-30_09-34-59](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2848/console).

PR #920, opened 2016-05-28
Use start values at first clock tick (ticket:3770)

rfranke:
This shall only apply to states of solved continuous equations.
We check for the existence of a solver method so far.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-05-28_22-54-36](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2845/console).

PR #931, opened 2016-06-01
New generic type ExpandableArray

lochel:
This introduces a new generic type to handle expandable arrays and an initial set of auxiliary functions:

- [x] **new**: Creates a new ExpandableArray of given type and initial size.
- [X] **get**: Returns the data of given array index.
- [x] **set**: Adds an element to given array position.
- [x] **add**: Adds an element to the first unused array position.
- [x] **delete**: Removes the element from the given position.
- [X] **update**: Overrides a given array position.
- [X] **clear**: Removes all elements.
- [x] **shrink**: Reduces the expandable array to a given size.
- [X] **dump**: Dumps the expandable array.

---

- [x] Clean up implementation
- [x] Use `MetaModelica.Dangerous` functions

### DIFF
--- a/Compiler/Util/ExpandableArray.mo
+++ b/Compiler/Util/ExpandableArray.mo
@@ -1,0 +1,247 @@
+/*
+ * This file is part of OpenModelica.
+ *
+ * Copyright (c) 1998-2014, Open Source Modelica Consortium (OSMC),
+ * c/o Linköpings universitet, Department of Computer and Information Science,
+ * SE-58183 Linköping, Sweden.
+ *
+ * All rights reserved.
+ *
+ * THIS PROGRAM IS PROVIDED UNDER THE TERMS OF GPL VERSION 3 LICENSE OR
+ * THIS OSMC PUBLIC LICENSE (OSMC-PL) VERSION 1.2.
+ * ANY USE, REPRODUCTION OR DISTRIBUTION OF THIS PROGRAM CONSTITUTES
+ * RECIPIENT'S ACCEPTANCE OF THE OSMC PUBLIC LICENSE OR THE GPL VERSION 3,
+ * ACCORDING TO RECIPIENTS CHOICE.
+ *
+ * The OpenModelica software and the Open Source Modelica
+ * Consortium (OSMC) Public License (OSMC-PL) are obtained
+ * from OSMC, either from the above address,
+ * from the URLs: http://www.ida.liu.se/projects/OpenModelica or
+ * http://www.openmodelica.org, and in the OpenModelica distribution.
+ * GNU version 3 is obtained from: http://www.gnu.org/copyleft/gpl.html.
+ *
+ * This program is distributed WITHOUT ANY WARRANTY; without
+ * even the implied warranty of  MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE, EXCEPT AS EXPRESSLY SET FORTH
+ * IN THE BY RECIPIENT SELECTED SUBSIDIARY LICENSE CONDITIONS OF OSMC-PL.
+ *
+ * See the full OSMC Public License conditions for more details.
+ *
+ */
+
+encapsulated uniontype ExpandableArray<T> "Implementation of an expandable array
+
+This provides a generic implementation of an expandable array. It basically
+behaves like an ordinary array, which means all elements can get accessed via
+index. When the array runs out of space, it get automatically resized. It is
+also possible to delete an element from any position."
+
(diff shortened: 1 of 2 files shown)


Compiler/Util/ExpandableArray.mo
```diff
@@ -0,0 +1,148 @@
...
+ *
+ * See the full OSMC Public License conditions for more details.
+ *
+ */
+
+encapsulated uniontype ExpandableArray<T> "Implementation of an expandable array"
+
+record ARRAY
+  Integer numberOfElements;
+  Integer size;
```
sjoelund: This is not quite safe. Should probably be array<Integer> numberOfElements, size; because right now you need to pass the expandable array as output everywhere and you will get really weird results for matchcontinue.

Compiler/Util/ExpandableArray.mo
```diff
@@ -0,0 +1,148 @@
...
+  array<Option<T>> data;
```
sjoelund: You may want to use the `MetaModelica.Dangerous` functions internally to avoid putting SOME(data) in here. Keeping `NONE()` at a position actually gives you nothing since this value should never be possible to access (you check bounds with `numberOfElements`).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-06-01_20-46-49](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2866/console).

perost:
What's wrong with Util/BaseVector.mo?

lochel:
@perost Util/BaseVector.mo is quite similar, but it does not support to delete elements from any position.

sjoelund:
@lochel - this package also does not support deleting an element at a position. Only clearing the entire table. Which for some reason is a linear-time operation instead of constant-time.

Compiler/Util/ExpandableArray.mo
lochel: @sjoelund This type allows to delete elements from any position. Therefore, it’s not enough to just check bounds with `numberOfElements`.

lochel:
@sjoelund The implementation is still not complete. There will be support to delete a certain element before this get merged.

Compiler/Util/ExpandableArray.mo
sjoelund: No it does not. The provided functions provide no such functionality. And if it does allow deleting arbitrary elements, the underlying datatypes are not suitable. I would suggest using the BaseVector or another efficient data-type and just put Option<T> into it to mark something "sort of deleted" instead of creating a type with a generic name that actually does something very specific.

Actually deleting something from the middle of a vector is quite complicated. I suspect what you want is a Deque with random access (like Qt::List).

Compiler/Util/ExpandableArray.mo
lochel: This is a good point. Probably I should do it like in Util/BaseVector.mo.
```Modelica
// The Vector type is an array of one VectorInternal, to make it completely mutable.
type Vector = array<VectorInternal>;
```

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-06-02_19-13-57](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2875/console).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-06-04_11-43-20](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2882/console).

hkiel:
@lochel @sjoelund Depending on the length of the used array it might make sense to store the index of the first free element in the ExpandableArray. Makes add() faster. (adding m elements then is of O(m)  and not O(m*n)!

sjoelund:
@hkiel, there is numberOfElements and capacity. This makes add as fast as it can be given that you cannot delete in the middle of the array. Which is something you should never do using a data-structure like this one. Better would probably be to just mark elements deleted and let them be deleted until a pass is made to clean up the array... There was nothing added that uses this data-type and I suspect it won't give good performance even for the considered use-case.

hkiel:
@sjoelund  looking at add() and delete() functions it is expected that you delete (mark as unused!) in the middle of the array, and adding looks from the first to the last slot for an empty place to put the new element in.
I also think this implementation does not perform very well...

hkiel:
@lochel @sjoelund Does it make sense to change ExpandableArray to use mutable<T> instead of array of size 1?

PR #953, opened 2016-06-07
simulation(..) return value #3953

bernhard-thiele:
Converts the Values.Value in function createSimulationResult into a
string value before returning it. As a result calling `simulation(..)`
in a mos script will return a string (as specified by the signature)
instead of a record.

@adrpo @sjoelund: This commit makes simulation(..) actually return a String (as indicated in the signature of ModelicaBuiltin.mo) instead of a record. However, I'm not sure what is really intended. Should be discussed.

### DIFF
--- a/Compiler/Script/CevalScriptBackend.mo
+++ b/Compiler/Script/CevalScriptBackend.mo
@@ -227,15 +227,17 @@ protected
   list<Values.Value> vals;
   list<String> fields;
   Boolean isTestType,notest;
+  Values.Value resRecord;
 algorithm
   resultValues := listReverse(inAddResultValues);
   //TODO: maybe we should test if the fields are the ones in simulationResultType_full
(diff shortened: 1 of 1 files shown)


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2016-06-08_09-52-27](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2905/).

sjoelund:
Do not convert the record into a string. You will break scripts very badly since they access the elements of the record. Like accessing `simRes.resultFile` to see if the simulation was successful. If you need to be able to convert a result to a string, create a function that is passed a record and returns a string.

lochel:
I guess the return type of the `simulate` command should get fixed, right?

bernhard-thiele:
Closed due to #1509.

PR #983, opened 2016-06-26
small memory optimizations

adrpo:
- free the array in List.mo
- return the input in SystemImpl__iconv if to == from
- signal that we don't want malloc_atomic in arrayCreateNoInit

### DIFF
--- a/Compiler/Util/List.mo
+++ b/Compiler/Util/List.mo
@@ -99,6 +99,7 @@ protected
 import MetaModelica.Dangerous.{listReverseInPlace, arrayGetNoBoundsChecking, arrayUpdateNoBoundsChecking, arrayCreateNoInit};
 import MetaModelica.Dangerous;
 import DoubleEndedList;
+import GC;
 
 public function create<T>
   "Creates a list from an element."
@@ -1044,6 +1045,7 @@ algorithm
 
     arrayUpdate(arr, i, false);
   end for;
+  GC.free(arr);
 end uniqueIntN;
 
 public function uniqueIntNArr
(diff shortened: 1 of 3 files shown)


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2016-06-26_13-45-24](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2966/).

Compiler/runtime/systemimpl.c
```diff
@@ -2230,6 +2230,7 @@ extern char* SystemImpl__iconv(const char * str, const char *from, const char *t
   iconv_t ic;
   int count;
   char *buf;
+  if (!strcmp(from, to)) return (char*)str;
```
sjoelund: This is wrong. iconv from UTF-8 to UTF-8 does the magical operation: verify that the input is in UTF-8.

adrpo: @sjoelund: yes, i got it. but can be moved further below, or we could check if the output is the same and free the memory.

sjoelund: Well, we could do that check always since usually things are ASCII 7-bit and won't need the allocated memory. It's also a bit of a problem since the external C interface will always make a copy of the string anyway (`external "MetaModelica"` would be nice to have).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-06-26_14-55-32](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2967/console).

Compiler/Util/List.mo
```diff
@@ -1044,6 +1045,7 @@ algorithm
 
     arrayUpdate(arr, i, false);
   end for;
+  GC.free(arr);
```
hkiel: Does this improve performance? There are some more places where temporary arrays could be free'd

adrpo: @hkiel: it make it easier for GC to reuse empty spaces so I would say it does.

PR #1005, opened 2016-07-05
do not use extrapolation for to big steps

wibraun:
- let the nonlinear solver use old values if the to extrapolated value is 5*times larger then step size.
- also fix minor spelling mistakes


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-07-05_14-01-27](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/2999/console).

PR #1006, opened 2016-07-05
add further daeMode option

wibraun:
This adds an additional option for the daeMode=[all,dynamic]. 
So far the dynamic option was default, with the option "all", all equations (also the output algebraic equations) are used by the integrator, this reduces the costs of the zero crossings evaluation, but increases the problem size further.

### DIFF
--- a/Compiler/SimCode/SimCodeUtil.mo
+++ b/Compiler/SimCode/SimCodeUtil.mo
@@ -241,6 +241,7 @@ protected
   BackendDAE.EquationArray bdaeModeEqns;
   Option<SimCode.JacobianMatrix> daeModeSP;
   Option<SimCode.DaeModeData> daeModeData;
+  SimCode.DaeModeConfig daeModeConf;
   list<SimCodeVar.SimVar> sensitivityVars;
   Integer countSenParams;
   list<tuple<Integer, Integer>> equationSccMapping, eqBackendSimCodeMapping;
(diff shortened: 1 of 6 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-07-05_15-10-59](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3000/console).

Compiler/SimCode/SimCodeUtil.mo
```diff
@@ -437,11 +438,12 @@ algorithm
     end if;
 
     // add residuals vars from DAE creation
-    if Flags.getConfigBool(Flags.DAE_MODE) then
+    if Flags.getConfigEnum(Flags.DAE_MODE)>1 then
       daeModeSP := createDaeModeSparsePattern(bdaeModeVars, bdaeModeEqns, shared, crefToSimVarHT);
       residualVars := rewriteIndex(residualVars, 0);
       algebraicVars := sortSimVarsAndWriteIndex(algebraicVars, crefToSimVarHT);
-      daeModeData := SOME(SimCode.DAEMODEDATA(daeEquations, daeModeSP, residualVars, algebraicVars));
+      daeModeConf := match Flags.getConfigEnum(Flags.DAE_MODE) case 2 then SimCode.ALL_EQUATIONS(); case 3 then SimCode.DYNAMIC_EQUATIONS(); end match;
```
hkiel: is it wise to have magic numbers 2 and 3 here???

wibraun: I'm open to other solutions, any proposal?
Actually, exactly in this case I think it's okay to handle with enumeration numbers, they aren't that magic,  but below in the c-code they get more magical, I don't like that solution that much, but it works ;).

PR #1015, opened 2016-07-08
Fix equation system partitioning performance

sjoelund:
There was a scalability issue for systems where many equations were
connected to the same variable - variables were not checked for
visited status so the same variable was visited multiple times.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-07-08_10-26-22](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3013/console).

sjoelund:
@casella 10 seconds slower on average to fix the worst case in your model

PR #1023, opened 2016-07-13
Avoid characters in file names that are illegal in Windows

dietmarw:
In https://trac.modelica.org/Modelica/ticket/2024 it was noticed that
the help files generated by this OMC script contained the under
Windows illegal characters '<' and '>'. This patch fixes that issue.

The characters are replaced by something that resambles URI encoding
but not quite since using actual "%3C" causes other issues when overly
clever browser revert "%3c" into "<" and hence the filename can not
be found by the browser. Therefore this patch uses "x3C" and "x3E".

### DIFF
--- a/Examples/GenerateDoc.mos
+++ b/Examples/GenerateDoc.mos
@@ -89,8 +89,12 @@ function filename \"Replace characters that mess with filesystems and shell expa
   input String str;
   output String ostr :=
         OpenModelica.Scripting.stringReplace(
-          OpenModelica.Scripting.stringReplace(str,\"/\",\"Division\"),
-          \"*\",\"Multiplication\")
+        OpenModelica.Scripting.stringReplace(
+        OpenModelica.Scripting.stringReplace(
+        OpenModelica.Scripting.stringReplace(str,\"/\",\"Division\"),
+        \"*\",\"Multiplication\"),
+        \"<\",\"x3C\"),
+        \">\",\"x3E\")
 ;
 end filename;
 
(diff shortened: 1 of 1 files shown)


adrpo:
Maybe we should address all the chars given here:
https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx
so we don't run into this issue anymore.

dietmarw:
I wanted to keep it simple and to not hit any false positives.

dietmarw:
Also we only need to address characters that are legal in Modelica class names and illegal on windows file system names. That overlap might even be small then the list.

adrpo:
@dietmarw: true, is fine like it is.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-07-13_17-49-44](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3027/console).

svorkoetter:
Just to clarify, is OM used to generate the files that ship with the MSL? Regarding Dietmar's comment above that, "we only need to address characters that are legal in Modelica class names and illegal on windows file system names", that would be all printable ASCII characters, would it not, since you can use quotes to allow otherwise illegal characters.

sjoelund:
Yes, I believe Dietmar has used this script to generate the MSL zip-file (the generated files are not in the git repo of MSL). And yes, all printable 7-bit ASCII characters are allowed in Modelica class names.

PR #1026, opened 2016-07-15
Do not pass around numRelations

sjoelund:
DoubleEndedList already stores the number of elements in the list so it
is a constant-time operation to access the length without passing it
around.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-07-15_12-43-02](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3031/console).

PR #1095, opened 2016-10-01
Import fmu model description

alash325:


adrpo:
you need to update your git working copy and redo this pull request, i already merged the previous code and it has conflicts with this one.

sjoelund:
You should not add revert commits to a pull request. You should remove the commit that is not needed (in this case, I guess `git checkout importFMUModelDescription && git reset --hard 8e6bf34 && git push --force`).

alash325:
I updated my  working copy but When I tried to create a new pull request it comes with the previous pull request which are already merged by Adrian. Anything I missed?

PR #1100, opened 2016-10-04
do not link some external library resources (user32,pthreads,X11)

vwaurich:


Compiler/SimCode/SimCodeFunctionUtil.mo
```diff
@@ -2123,6 +2123,27 @@ algorithm
     case Absyn.STRING("lapack") then ({},{});
     case Absyn.STRING("Lapack") then ({},{});
 
+    //pthreads is already linked under windows
+    case Absyn.STRING(str as "pthreads")
```
sjoelund: `case Absyn.STRING(str as "pthreads") guard System.os()=="Windows_NT"` looks nicer, I guess :)

vwaurich: True. Thanks for the review.

PR #1110, opened 2016-10-07
Handle errors during FMI initialization

sjoelund:


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2016-10-07_08-16-20](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3192/).

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-10-02_11-49-39](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4606/).

SimulationRuntime/fmi/export/fmi2/fmu2_model_interface.c
```diff
@@ -390,6 +394,11 @@ fmi2Component fmi2Instantiate(fmi2String instanceName, fmi2Type fmuType, fmi2Str
   /* allocate memory for state selection */
   initializeStateSetJacobians(comp->fmuData, comp->threadData);
 #endif
+  fmi2callbacksOK = 1;
+  MMC_CATCH_INTERNAL(mmc_jumper)
+  if (!fmi2callbacksOK) {
+    return NULL;
+  }
```
adeas31: This looks a bit strange. The if condition will always fail since `fmi2callbacksOK` is set to 1 unless you are doing something in `MMC_CATCH_INTERNAL` which is of course not the case since `fmi2callbacksOK` is local.

sjoelund: It's not weird. It's setjmp/longjmp. The assignment to fmi2callbacksOK doesn't occur if an error is triggered.

sjoelund: OK, it's a bit wrong since it returns before restoring the jump buffer. But a bigger problem is that when OM imports the FMU, we use thread-local storage for the threadData and we need to have separate threadData for the importing model... This seems to be what causes the crashes. Maybe we need to make the simulation check for existing threadData and re-use that...

PR #1117, opened 2016-10-11
Fix use of scale factor in unit conversion, ticket:4031

rfranke:
This bug showed up with the newly introduced degF.

Also re-introduce ms, because OMEdit uses it for scaling of plots
and millisecond is globally unique.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-10-11_07-43-50](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3203/console).

PR #1118, opened 2016-10-11
Do not link expat with omc

sjoelund:
Expat is linked in via FMIL, but we do not need FMIL to bootstrap
so it should not be hard-coded.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-10-11_08-52-29](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3204/console).

PR #1123, opened 2016-10-12
Link against fewer libraries when bootstrapping

sjoelund:
There is no need to link against CORBA, reading of simulation results,
FMI parsing, HPCOM, etc during bootstrapping stage 1 so these are no
longer linked during stage 1.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-10-12_09-19-11](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3212/console).

PR #1137, opened 2016-10-19
Structural parameters as HashSet instead of Table

sjoelund:
The FCore.Cache stores the structural parameters as a HashTable, but
does not use the value. This changes the structure to a HashSet
instead.

Note: This also fixes a bug in a model where the HashTable gets
corrupted (probably due to matchcontinue use). The structure should
possibly be changed to an AvlSet instead (or HashTable/HashSet
changed to be fully mutable).

### DIFF
--- a/Compiler/FFrontEnd/FCore.mo
+++ b/Compiler/FFrontEnd/FCore.mo
@@ -45,10 +45,10 @@ import BaseAvlTree;
 import DAE;
 import SCode;
 import Prefix;
-import HashTable;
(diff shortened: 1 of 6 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-10-19_08-08-38](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3238/console).

PR #1143, opened 2016-10-19
Correct handling of parameters

ptaeuber:
This commit corrects the improper and inconsistent handling
of parameters by implementing what is listed below:

- Only evaluate structural parameters and parameters with
  annotation(Evaluate=true) during compile time by default.
  Final parameters and protected parameters are not evaluated
  any more.
  (see [ticket:3211](https://trac.openmodelica.org/OpenModelica/ticket/3211), [ticket:4027](https://trac.openmodelica.org/OpenModelica/ticket/4027))

- Many different preOpt modules are merged to one preOpt module
  "evaluateParameters" with different flags instead
  -> To calculate additional parameters the following flags are introduced:
     --evaluateFinalParameters=\<true|false\> (default: false)
     --evaluateProtectedParameters=\<true|false\> (default: false)

  -> To activate/deactivate the replacement of the calculated parameters
     in the DAE use:
     --replaceEvaluatedParameters=\<true|false\> (default: true)

- Mark structural parameters deciding the if-branch final
  (see ticket [ticket:4053](https://trac.openmodelica.org/OpenModelica/ticket/4053))

- Structural parameters in instances are detected correctly
  (see [ticket:4059](https://trac.openmodelica.org/OpenModelica/ticket/4059))

- Propagate the final attribute to parameters structural parameters
  depend on
  (see [ticket:4052](https://trac.openmodelica.org/OpenModelica/ticket/4052))

- Mark parameters final if used for calculation of other parameters,
  because otherwise it leads to inconsistency when changing the value
  after compilation.
  Also replace the evaluated expressions because they are already evaluated
  and not changeable anymore.
  (see [ticket:4027](https://trac.openmodelica.org/OpenModelica/ticket/4027), [ticket:3211](https://trac.openmodelica.org/OpenModelica/ticket/3211))

- Make final, structural and protected parameters unchangeable
  -> Write final parameters to _08bnd.c
  -> Make use of the isValueChangeable-flag
  (see [ticket:4060](https://trac.openmodelica.org/OpenModelica/ticket/4060), [ticket:3211](https://trac.openmodelica.org/OpenModelica/ticket/3211), [ticket:4027](https://trac.openmodelica.org/OpenModelica/ticket/4027))

### DIFF
--- a/Compiler/BackEnd/BackendDAEUtil.mo
+++ b/Compiler/BackEnd/BackendDAEUtil.mo
@@ -4978,6 +4978,14 @@ algorithm
         (e,_) = Expression.traverseExpBottomUp(e, replaceVartraverser, vars);
       then (e, vars);
 
+    case (DAE.CREF(componentRef=cr),vars)
+      equation
+        (v::_,_) = BackendVariable.getVar(cr,vars);
+        true = BackendVariable.varFixed(v);
+        e = BackendVariable.varBindExpStartValue(v);
+        (e,_) = Expression.traverseExpBottomUp(e, replaceVartraverser, vars);
+      then (e, vars);
+
     else (inExp,inVars);
 
   end matchcontinue;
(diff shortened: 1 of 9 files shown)


ptaeuber:
Among others the Backend module "evaluateParameters" evaluates structural parameters, which is only
a workaround, since this is already done in the Frontend but the values get lost.

There are two models in the testsuite which use one of the flags to evaluate also final parameters
or protected parameters, otherwise they would fail:

`/simulation/libraries/msl32/Modelica.Electrical.Machines.Examples.Transformers.Rectifier12pulse.mos`
  --> Different simulation results without evaluating final parameters.
      The cause of it should be investigated as soon as possible, since it breaks the coverage
      test for this model.
`/openmodelica/fmi/ModelExchange/1.0/Modelica_Mechanics_MultiBody_Examples_Elementary_DoublePendulum.mos`
  --> Fails to build the _me_FMU.mo without evaluating the protected parameters.
      Has something to do with the string parameters.
      
The commit might influence the whole coverage test greatly since less parameters
are evaluated and the systems can differ from the ones before and the solvers might react
in a sensitive way. Nevertheless, it is the correct way to not evaluate final and protected
parameters by default.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-10-20_23-12-13](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3253/console).

PR #1151, opened 2016-10-21
Fix order of clocked equations

sjoelund:
Before, the IDs given to clocked equations was in the middle of the
regular equations, but it needs to come after (or before).


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-10-21_14-14-07](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3258/console).

PR #1154, opened 2016-10-24
BackendDAE.ASSIGN can hold tuples as lhs

vwaurich:
needed to fix packager in M_DD
function calls which make use of external functions are not allowed to be expanded since this will end up in multiple function calls which leads to unintended behaviour


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2016-10-24_10-47-23](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3273/).

PR #1160, opened 2016-10-24
Filter protected and hidden in Cpp runtime, ticket:4089

rfranke:
The former global setting ResultsOutput {true, false} is replaced
with enum EmitResults {EMIT_ALL, EMIT_PUBLIC, EMIT_NONE}.
The default value is EMIT_ALL (cf. former true).

OMCFactory uses the default value EMIT_PUBLIC and introduces the
new option --emit-results with the values all, public or none.
The OMEdit option `-emit-protected` is tranlated to `--emit-results all`.

The actual filtering simply skips values in XmlPropertyReader. This works
out as protected values are not needed from init.xml because they are
calculated in the generated code.

### DIFF
--- a/Compiler/Template/CodegenCpp.tpl
+++ b/Compiler/Template/CodegenCpp.tpl
@@ -5876,9 +5876,9 @@ case SIMCODE(modelInfo = MODELINFO(__),makefileParams = MAKEFILE_PARAMS(__))  th
    {
       #if !defined(FMU_BUILD)
         #if defined(__vxworks)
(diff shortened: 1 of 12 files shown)


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2016-10-24_22-46-52](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3284/).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-10-24_23-25-48](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3285/console).

PR #1169, opened 2016-10-26
Override protected with HideResult annotation, ticket:4089

rfranke:
This treats protected variables with annotation(HideResult=false)
like public variables during model translation.
In particular such parameters can be changed after model translation.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-10-26_20-26-34](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3310/console).

PR #1202, opened 2016-11-04
Handle comments moving around

sjoelund:
This fixed ticket:4065, where comments move around and the diff
algorithm gets confused. It works by moving comment tokens from one
scope to another. By getting closer to the original point, the
existing diff algorithm (often) manages to get the comment into the
correct position.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-11-04_10-18-16](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3374/console).

PR #1220, opened 2016-11-10
Do not crash when reading a parameter

sjoelund:
The filterSimulationResults API expects only variables when converting
to CSV. This adds a check to see if a parameter was passed instead.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-11-10_15-56-48](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3406/console).

PR #1234, opened 2016-11-18
Make test utils able to clean more generated files

sjoelund:
This is necessary since Cppruntime statically links its runtime system
with the simulation. It caused the library testing server to have 50GB
of unused files lying around. It is less crucial for the C runtime since
the executable is always cleaned (this is an so-file for cppruntim
which was not cleaned up), and because the objects were small (few kB
instead of 10MB).

### DIFF
--- a/Compiler/SimCode/SimCodeMain.mo
+++ b/Compiler/SimCode/SimCodeMain.mo
@@ -54,6 +54,7 @@ import SimCode;
 
 // protected imports
 protected
+import AvlSetString;
 import BackendDAECreate;
 import Builtin;
 import ClockIndexes;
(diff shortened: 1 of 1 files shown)


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2016-11-18_13-37-41](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3429/).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-11-18_14-50-14](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3430/console).

PR #1236, opened 2016-11-21
Type checking updates for the new fron-end

mahge:
Filling up the missing cases for equations and algorithms.
  Updated binary operation checks.

### DIFF
--- a/Compiler/NFFrontEnd/NFTypeCheck.mo
+++ b/Compiler/NFFrontEnd/NFTypeCheck.mo
@@ -609,7 +609,231 @@ algorithm
   end try;
 end checkBinaryOperation;
 
+
 public function checkBinaryOperationArrays
+  "mahge:
+  Type checks binary operations involving arrays. This involves more checks than
+  scalar types. All normal operations as well as element wise operations involving
+  arrays are handled here."
+  input DAE.Exp inExp1;
+  input DAE.Type inType1;
+  input DAE.Operator inOp;
+  input DAE.Exp inExp2;
+  input DAE.Type inType2;
+  output DAE.Exp outExp;
+  output DAE.Type outType;
+protected
+  Boolean isArray1,isArray2, bothArrays;
+  Integer nrDims1, nrDims2;
+algorithm
+
+  nrDims1 := getNrDims(inType1);
+  nrDims2 := getNrDims(inType2);
+  isArray1 := nrDims1 > 0;
+  isArray2 := nrDims2 > 0;
+  bothArrays := isArray1 and isArray2;
+
+  (outExp, outType) := match inOp
+    local
+      DAE.Exp exp1,exp2;
+      DAE.Type ty1,ty2, arrtp, ty;
+      DAE.Dimensions dims;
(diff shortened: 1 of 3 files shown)


Compiler/NFFrontEnd/NFTypeCheck.mo
```diff
@@ -621,6 +622,229 @@ public function checkBinaryOperationArrays
...
+          ty2 := underlyingType(inType2);
+          // TODO: one of the exps is array but its type is now simple.
+          (exp1,exp2,ty) := matchTypeBothWays(inExp1,ty1,inExp2,ty2);
+
+          outType := promoteLeft(ty,dims);
+          outExp := DAE.BINARY(exp1, DAE.MUL_ARRAY_SCALAR(outType), exp2);
+
+        elseif (nrDims1 == 1 and nrDims2 == 1) then
+          {N1} := getArrayTypeDims(inType1);
+          {N2} := getArrayTypeDims(inType1);
```
sjoelund: Shouldn't it be type2?

Compiler/NFFrontEnd/NFTyping.mo
```diff
@@ -445,6 +609,33 @@ algorithm
...
+    (e1, ty1, var1) := typeExp(absExp, NFComponent.DEFAULT_SCOPE, component, info);
+    typedOpExp := SOME(e1);
+    opTy := SOME(ty1);
+    OpVarity := SOME(var1);
+  else
+    typedOpExp := NONE();
+    opTy := NONE();
+    OpVarity := NONE();
+  end if;
+end typeExpOption;
```
sjoelund: Seems like an odd function to add (it is slow; and anyway unused)

mahge: This will be needed in some places where we have optional expressions, e.g., for loop ranges. It is unused now because for loops are skipped until iterator scopes are handled by Per. 
May I ask why it is considered slow? What is the alternative? feel free to change it.

Compiler/NFFrontEnd/NFTypeCheck.mo
mahge: Ya this should be inType2. Probably a couple more of errors like this exist. Will filter them out when we have tests in place. Will fix this for now.

Compiler/NFFrontEnd/NFTyping.mo
sjoelund: You don't need it if you match on SOME() / NONE() in the pattern for the for-loop. Then you can avoid creating 3 SOME-nodes.

mahge: Removed.

mahge:
> What about operator overloading?
you mean user defined operator overloading? That is a long shot right now. Too many things missing for it.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-11-22_13-40-19](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3444/console).

PR #1241, opened 2016-11-21
Fix the check for final modification.

perost:
- Moved the final attribute from NFMod.Modifier to NFBinding.Binding,
  so that an element can be declared final without having a binding.
- Update the check for merging two modifiers where the inner is final.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-11-21_16-43-47](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3442/console).

PR #1244, opened 2016-11-23
New instantiation improvements.

perost:
- Fixed scoping when instantiating inherited components.
- Fixed userdefined types so they can be used more than once.
- Fixed userdefined type which extends from userdefined type.
- Added error message for combining extends of builtin type with other
  elements.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-11-23_21-57-08](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3446/console).

PR #1274, opened 2016-12-04
NFInst improvements.

perost:
- Ported state machine for checking lookup restrictions from old new
  inst to new inst.
- Many other lookup improvements.
- Fixed cloning of instances so that modifiers don't stick to classes.
- Other minor instantiation fixes.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-12-04_13-32-01](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3499/console).

PR #1291, opened 2016-12-09
unify the way we handle version in Windows / Linux

adrpo:
- add OpenModelica prefix to the CONFIG_REVISION in omc_config.h
- removed OpenModelica prefix from Windows generated revision.h

### DIFF
--- a/Compiler/runtime/omc_config.h
+++ b/Compiler/runtime/omc_config.h
@@ -155,9 +155,9 @@
 #endif /* #if !defined(MSYS2_AUTOCONF) && (defined(__MINGW32__) || defined(_MSC_VER)) */
 
 #ifdef CONFIG_REVISION
-#define CONFIG_VERSION CONFIG_REVISION
+#define CONFIG_VERSION "OpenModelica " CONFIG_REVISION
 #else
-#define CONFIG_VERSION "unknown"
+#define CONFIG_VERSION "OpenModelica unknown"
 #endif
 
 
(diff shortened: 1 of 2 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-12-09_02-42-49](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3529/console).

sjoelund:
+#define CONFIG_VERSION "OpenModelica " CONFIG_REVISION will not duplicate the Linux behaviour. Linux displays either "OpenModelica ..." or "OMCompiler ..." depending on which version/hash is being used.

casella:
Does this fix [#3745](https://trac.openmodelica.org/OpenModelica/ticket/3745)? If so, please close the ticket.

adrpo:
I think it needs more work. I guess in some cases we might double the "OpenModelica " prefix. I'll do some more checking.

sjoelund:
And in some cases Windows builds might display OpenModelica for an OMCompiler hash?
On Dec 10, 2016 00:44, Adrian Pop <notifications@github.com> wrote:I think it needs more work. I guess in some cases we might double the "OpenModelica " prefix. I'll do some more checking.

adrpo:
So how should we deal with this? We don't put any prefix on the thing we get from git and we add it inside the programs: omc, omedit?

adrpo:
Also we need to keep in mind that we use the version number in some scripts (Hudson jobs) which expect no space in it.

PR #1295, opened 2016-12-12
Separated linear solver and nonlinear solver in two classes classes

niklwors:


niklwors:
merge conflicts has to be resolved first

PR #1314, opened 2016-12-16
Add newline after annotation in AST dump - list()

sjoelund:
This newline is so OMEdit gets code folding by default. Each top-
level modification in the annotation gets its own line.

Also now handles newline in annotation String literals better (by not
converting Text to string in Susan, which messes with indentation).


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-12-16_08-48-40](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3578/console).

PR #1316, opened 2016-12-16
Copy entire array when a dimension is unknown

sjoelund:
Only copying the data only works if the array has allocated the
correct amount of memory.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-12-16_11-21-25](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3581/console).

PR #1320, opened 2016-12-16
small fixes for Windows building

adrpo:
- we don't have mico anymore, use use omniOrb
- set global roots to zero by default
- if we find OpenCL.dll in Windows generate libOpenCL.a so that ParModelica can be used


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-12-17_00-16-47](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3587/console).

PR #1325, opened 2016-12-18
new frontend improvements

adrpo:
- copy sections from inherited classes to the instantiated class in NFClass.instExpandedClass
- fix some typing issues
- handle "pre"
- do some pour man's evaluation on the dae
- handle FOR equations with simple ranges
- handle WHEN equations
- transform SCode.ENUMERATION in a class with parts
- whith this about 37 models from ScalableTestSuite are flattening

### DIFF
--- a/Compiler/NFFrontEnd/NFClass.mo
+++ b/Compiler/NFFrontEnd/NFClass.mo
@@ -136,14 +136,62 @@ uniontype Class
     input array<InstNode> components;
     input Class expandedClass;
     output Class instancedClass;
+  protected
+    list<Equation> eqs = {}, eqs1 = {}, eqs2 = {};
+    list<Equation> ieqs = {}, ieqs1 = {}, ieqs2 = {};
+    list<list<Statement>> algs = {}, algs1 = {}, algs2 = {};
+    list<list<Statement>> ialgs = {}, ialgs1 = {}, ialgs2 = {};
   algorithm
     instancedClass := match expandedClass
       case EXPANDED_CLASS()
-        then INSTANCED_CLASS(expandedClass.elements, components,
-          expandedClass.equations, expandedClass.initialEquations,
-          expandedClass.algorithms, expandedClass.initialAlgorithms);
+        algorithm
+          eqs1 := expandedClass.equations;
+          ieqs1 := expandedClass.initialEquations;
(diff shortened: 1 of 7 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-12-18_06-20-56](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3592/console).

Compiler/NFFrontEnd/NFClass.mo
```diff
@@ -136,15 +136,63 @@ uniontype Class
...
+          ialgs1 := expandedClass.initialAlgorithms;
+
+          // append the sections from the inherited classes
+          for ext in expandedClass.extendsNodes loop
+            (eqs2, ieqs2, algs2, ialgs2) := collectInherited(ext, eqs2, ieqs2, algs2, ialgs2);
+          end for;
+          eqs := listAppend(eqs2, eqs1);
+          ieqs := listAppend(ieqs2, ieqs1);
+          algs := listAppend(algs2, algs1);
+          ialgs := listAppend(ialgs2, ialgs1);
```
perost: I don't think we want to do this here. Inherited sections still need to be instantiated using the scope of the extends node, otherwise you'll be able to find any name in the inheriting class when doing lookup. I don't really see any need to copy sections anywhere except for the flattened DAE.

adrpo: If i don't do this the DAE gets no equations from the inherited classes. Maybe it should be done sompeplace else but it needs to be done.

adrpo: The extendsNodes go away when going from EXPANDED_CLASS to INSTANCED_CLASS so you cannot access them anymore.

perost:
Yes, because I haven't implemented anything regarding inheritance of sections. The extends nodes should be added to the instanced class, and then they can be flattened and so on. Then we don't need to append sections in the classes at all, just add them to the DAE during flattening. Right now we don't handle extends correctly after the instantiation, so e.g. a model such as this actually passes:
```modelica
model A
  Real x = y;
end A;

model B
  Real y = z;
end B;

model C
  Real z;
  extends A;
  extends B;
end C;
```

PR #1338, opened 2016-12-21
Add some sanity checks for threads

sjoelund:
- Check pthread status codes
- Check that we executed all tasks
- Check that all returned types are not mixed boxed/immediate (usual
  case is the values contain a NULL value when some data was expected)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2016-12-21_13-17-26](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3613/console).

PR #1341, opened 2017-01-04
Move external object constructors to initialization

lochel:
see [ticket:3446](https://trac.openmodelica.org/OpenModelica/ticket/3446#comment:8)

- [x] This need to be tested with some examples where parameters depend on external objects
- [x] @adrpo wrote: I guess one needs to have a look at the FMI stuff too, so that the order is changed there too.
- [x] Clean up code generation

### DIFF
--- a/Compiler/BackEnd/Initialization.mo
+++ b/Compiler/BackEnd/Initialization.mo
@@ -806,7 +806,14 @@ algorithm
   allParameterEqns := BackendEquation.emptyEqns();
   otherVariables := BackendVariable.emptyVars();
   (allParameters, allParameterEqns, otherVariables) := BackendVariable.traverseBackendDAEVars(dae.shared.globalKnownVars, selectParameter2, (allParameters, allParameterEqns, otherVariables));
-  (allParameters, allParameterEqns, otherVariables) := BackendVariable.traverseBackendDAEVars(dae.shared.externalObjects, selectParameter2, (allParameters, allParameterEqns, otherVariables));
+
+  // FIXME: This if clause is a workaround. Remove it as soon as cpp runtime is
+  // adapted to handle external objects together with parameters.
+  // see ticket:3446
+  if not stringEq(Config.simCodeTarget(), "Cpp") then
+    (allParameters, allParameterEqns, otherVariables) := BackendVariable.traverseBackendDAEVars(dae.shared.externalObjects, selectParameter2, (allParameters, allParameterEqns, otherVariables));
+  end if;
+
   nParam := BackendVariable.varsSize(allParameters);
 
   if nParam > 0 then
--- a/Compiler/BackEnd/Initialization.mo
+++ b/Compiler/BackEnd/Initialization.mo
@@ -780,7 +780,7 @@ protected function selectInitializationVariablesDAE "author: lochel
   This function wraps selectInitializationVariables.
   All primary parameters get removed from the dae."
   input output BackendDAE.BackendDAE dae;
-  output BackendDAE.Variables outVars;
+  output BackendDAE.Variables outInitVars;
   output list<BackendDAE.Var> outPrimaryParameters = {};
   output list<BackendDAE.Var> outAllPrimaryParameters = {};
 protected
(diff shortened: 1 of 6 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-01-04_13-54-11](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3615/console).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-01-04_14-15-31](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3616/console).

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-01-09_13-19-38](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3633/).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-01-09_13-45-43](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3634/console).

lochel:
@rfranke This is what I've changed for the c runtime. I would also like to fix it for the cpp runtime before merging. Maybe you can help me with that.

rfranke:
The Cpp runtime constructs and destructs external objects at one place, no matter if it generates a simulator or an FMU. This commit 05cfc469d7706eccc61e045a393df2d740f20948 touched the construction of external objects. It basically had moved the call of `constructExternalObjects()` to the initialization of bound parameters (but without sorting) and introduced the flag `_constructedExternalObjects` to make sure they are only constructed once.

Do you mean that `constructExternalObjects()` and  the flag `_constructedExternalObjects` should be removed?
Should the call to `destructExternalObjects()` stay in the generated code?

lochel:
> Do you mean that `constructExternalObjects()` and the flag `_constructedExternalObjects` should be removed?

Yes, `constructExternalObjects()` should be removed, since the external objects are now also part of the bounded parameter list. That means the code generation of bounded parameters need to be adapted to also handle external objects. I don't know if `_constructedExternalObjects` is needed.

> Should the call to destructExternalObjects() stay in the generated code?

Yes, the destructors should just stay as they are.

rfranke:
What do you do if a model gets initialized multiple times, e.g. after fmi2Reset()? The current implementation in the Cpp runtime calls `destructExternalObjects()` during initialization of free parameters if `_constructedExternalObjects` is true.

lochel:
I don't think that this is already handled in the c runtime. However, it seems reasonable to keep `_constructedExternalObjects`to handle that case.

Do we have a test for fmi, `fmi2Reset()`, and external objects?

rfranke:
I'm afraid we have no tests for "non-standard" calling sequences.

sjoelund:
If you `fmi2Reset()`, you need to destroy the external objects and construct them again during initialization since parameters may change

rfranke:
This is why the Cpp runtime destructs already existing external objects during (the next) initialization of free parameters.

You might also merge the pull request and we see how to get rid of the workaround for Cpp afterwards.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-01-10_08-41-50](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3640/console).

lochel:
> If you fmi2Reset(), you need to destroy the external objects and construct them again during initialization since parameters may change

I see your point. Can one provide a test case for this? Then, I will take care of fixing this for the c runtime.

sjoelund:
I can't, because we don't have an FMI simulator that works in that way (we never call fmi2Reset as far as I know). External tools might though...

lochel:
@rfranke I’ve tried to adapt the cpp runtime myself and failed. However, it should be fairly simple. Basically, the function that calls all the external object constructors need to get removed and the code generation of bounded parameters need to be extended to handle also external objects. That's it.

PR #1345, opened 2017-01-09
Introduce subactive evaluation mode for clocked equations

rfranke:
Subactive evaluation means that discrete states are kept constant
while all other discrete variables are updated. This is useful for
numerical analysis, like calculation of partial derivatives.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-01-09_06-45-20](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3631/console).

PR #1348, opened 2017-01-09
use memset instead of non-standard bzero

adrpo:
bzero is non standard and not supported on msys2/mingw, use memset instead


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-01-09_22-52-21](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3638/console).

PR #1352, opened 2017-01-11
Remove special treatment of string return args, ticket:3446

rfranke:
The removed code assigned an uninitialized temporary strVar
to the output of the function. It does not appear needed because
std::string can be assigned like any other variable.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-01-11_09-08-58](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3649/console).

PR #1379, opened 2017-01-17
Fix code generation where `&` is used on an rvalue

sjoelund:
Added a daeExpReturnLValue function to Susan which converts an rvalue to
an lvalue when the code requires it. Can also be used when you want to
cache the expression and use its value twice (without evaluating the
expression twice).

This fixes ticket:4212.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-01-17_16-10-14](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3698/console).

PR #1381, opened 2017-01-18
Make linear systems not depend on nonlinear

sjoelund:
Make shared debugging routines static inline functions. This helps
compiling source FMUs that contain linear systems but no non-linear
systems.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-01-18_13-10-21](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3701/console).

PR #1389, opened 2017-01-19
New index reduction method "none"

lochel:
This method can be selected using --indexReductionMethod=none.
It will not handle higher index problems at all and just fail in case to get proper error messages if no higher index problem is expected.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-01-19_09-34-46](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3714/console).

casella:
@lochel can you please quickly elaborate on the difference between `uode` and `none`?
Thanks!

lochel:
@casella The short answer is that `none` is just no index-reduction at all. It causes the matching process to fail as soon as it detects a structural singular system.

Whereas `uode` will try to handle structural singularities by replacing algebraic dependencies between states with their derivatives.

I guess this should be updated in the documentation together with some easy example (like pendulum) and a comparison of all index reduction methods.

PR #1396, opened 2017-01-21
Allow fmi2SetInteger in continuous-time mode

lochel:
see [ticket:4192](https://trac.openmodelica.org/OpenModelica/ticket/4192)

### DIFF
--- a/SimulationRuntime/fmi/export/fmi2/fmu2_model_interface.c
+++ b/SimulationRuntime/fmi/export/fmi2/fmu2_model_interface.c
@@ -658,7 +658,7 @@ fmi2Status fmi2SetReal(fmi2Component c, const fmi2ValueReference vr[], size_t nv
 fmi2Status fmi2SetInteger(fmi2Component c, const fmi2ValueReference vr[], size_t nvr, const fmi2Integer value[]) {
   int i;
   ModelInstance *comp = (ModelInstance *)c;
-  if (invalidState(comp, "fmi2SetInteger", modelInstantiated|modelInitializationMode|modelEventMode))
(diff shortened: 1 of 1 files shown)


adeas31:
`fmi2SetInteger` shouldn't be allowed in continuous-time mode. See FMI specification page 85.

lochel:
It should be allowed to call `fmi2SetInteger` between two `fmi2DoStep` calls for co-simulation right? Therefore, it is necessary to allow that call in continuous-time mode for co-simulation. I will check how I can check more accurately if it is allowed to call `fmi2SetInteger`.

sjoelund:
How to check it: Implement the state machine defined in the FMI specification. It's not very many states...

lochel:
> How to check it: Implement the state machine defined in the FMI specification. It's not very many states...

I think there is already something like that implemented. But the current implementation does only consider the calling sequence of model exchange c functions and not the calling sequence of co-simulation c functions. I will investigate this a bit more.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-06-09_11-33-35](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4304/console).

adeas31:
This is not the fixed that we discussed, right?
It will now violates the model exchange fmus. We could check for fmu type and have different state checks based on it.

lochel:
Sure, we have to do that. But it seems that no one is working on the FMU export and we pretend to support ME and CS. This is now a blocker for CS for half a year!

lochel:
Now, I've improved the changes according to our discussion: #1674

PR #1397, opened 2017-01-21
Don't return success for unimplemented functions

lochel:
Or is there a good reason for returning `fmi2OK`?


adeas31:
No there is no good reason. Perhaps we should also add a message if a function is not supported by OpenModelica.

adeas31:
Better to use logger instead of `errorStreamPrint`.

lochel:
That's true - I will change that.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-01-22_17-26-51](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3735/console).

PR #1414, opened 2017-01-25
Add support for some synchronous elements

bernhard-thiele:
@perost Is this about the style which should be used for adding functions to the new front-end? Improvement suggestions?

### DIFF
--- a/Compiler/NFFrontEnd/NFFunc.mo
+++ b/Compiler/NFFrontEnd/NFFunc.mo
@@ -143,13 +143,11 @@ algorithm
        SCode.COMMENT(NONE(), NONE()),
        NONE(),
        info), scope);
-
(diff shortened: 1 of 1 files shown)


rfranke:
What is the rationale for "some" in the titile of the pull request?
Why not support all (about twice as many oparators than implemented so far)?

bernhard-thiele:
Because I want to understand what is good or bad style and what I can improve. Of course, the goal is to support everything finally.

adrpo:
@bernhard-thiele: for now this is OK. We might change it a bit later but  it should not need too much adaptation.

bernhard-thiele:
@adrpo OK, I will close this pull request and open a new one since I now have a few more operators which can be merged

PR #1422, opened 2017-01-26
Only print we remove variables actually removed

sjoelund:
Fix for the refactoring script getting confused because the notifications say we removed a variable, but the compiler actually does not.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-01-26_16-39-07](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3803/console).

PR #1423, opened 2017-01-26
partial implementation of class extends

adrpo:
- the class extends changes are not applied to the baseclass, only the local scope
- check if this is the correct way to implement this, or we actually need to replace in the base class

### DIFF
--- a/Compiler/NFFrontEnd/NFInst.mo
+++ b/Compiler/NFFrontEnd/NFInst.mo
@@ -59,12 +59,14 @@ protected
 import Array;
 import Error;
 import Flatten = NFFlatten;
+import Global;
 import InstUtil = NFInstUtil;
 import List;
 import Lookup = NFLookup;
 import MetaModelica.Dangerous;
 import Typing = NFTyping;
 import ExecStat.{execStat,execStatReset};
+import System;
 
 public
 function instClassInProgram
@@ -431,6 +433,13 @@ algorithm
       then
         node;
 
+    case SCode.CLASS(classDef = cdef as SCode.CLASS_EXTENDS())
+      algorithm
+        // get the already existing classes with the same name
+        print("Got class extends: " + def.name + "\n");
+      then
+        fail();
+
     else
       algorithm
         assert(false, getInstanceName() + " got unknown class");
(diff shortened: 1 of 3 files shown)


adrpo:
@perost: any comments about this?

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-01-26_22-50-36](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3805/console).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-01-26_23-36-18](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3806/console).

Compiler/NFFrontEnd/NFInst.mo
```diff
@@ -732,9 +741,31 @@ function addInheritedElements2
...
+  ClassTree.Entry idExisting;
+  InstNode nodeExisting;
 algorithm
   scope := match id
-    case ClassTree.Entry.CLASS() then ClassTree.add(inScope, name, id);
+    case ClassTree.Entry.CLASS()
+      algorithm
+        // check if class exists!
+        try
+          (idExisting as ClassTree.Entry.CLASS(nodeExisting)) := ClassTree.get(inScope, name);
```
perost: ClassTree.add (i.e. AvlTreeBase.add) takes a fourth argument which is a function called in the case of a collision. If you use that instead you don't incur the cost of a lookup for each class.

Compiler/NFFrontEnd/NFInst.mo
```diff
@@ -1388,5 +1419,155 @@ algorithm
...
+      list<Absyn.NamedArg> clats;
+      list<tuple<SCode.Element, DAE.Mod, Boolean>> rest;
+      tuple<SCode.Element, DAE.Mod, Boolean> first;
+      SCode.Mod mods, derivedMod;
+      DAE.Mod mod1,emod;
+      SourceInfo info1, info2;
+      Boolean b;
+      SCode.Attributes attrs;
+      Absyn.TypeSpec derivedTySpec;
+
```
perost: From what I can see there's a lot of unneccesary code here. You don't seem to actually change the base class (and you shouldn't need to), so it's just taken apart and then reassembled unchanged. There also doesn't seem to be any need for two different cases, since it doesn't matter what kind the base class is.

Compiler/NFFrontEnd/NFInst.mo
```diff
@@ -1388,5 +1419,155 @@ algorithm
...
+        name2 = inBaseEltName;
+
+        SCode.CLASS(_,prefixes2,encapsulatedPrefix2,partialPrefix2,restriction2,SCode.DERIVED(derivedTySpec, derivedMod, attrs),comment2,info2) = cl;
+
+        SCode.CLASS(_, prefixes1, encapsulatedPrefix1, partialPrefix1, restriction1, classExtendsCdef, comment1, info1) = classExtendsElt;
+        SCode.CLASS_EXTENDS(_,mods,SCode.PARTS(els1,nEqn1,inEqn1,nAlg1,inAlg1,inCons1,_,externalDecl1)) = classExtendsCdef;
+
+        classDef = SCode.DERIVED(derivedTySpec, derivedMod, attrs);
+        compelt = SCode.CLASS(name2,prefixes2,encapsulatedPrefix2,partialPrefix2,restriction2,classDef,comment2,info2);
+        vis2 = SCode.prefixesVisibility(prefixes2);
```
perost: This doesn't seem correct to me, I would assume that in a class extends the extends clause is always public.

PR #1438, opened 2017-01-30
Append underscore to names in Cpp functions

rfranke:
Remove special treatment for a subset of C++ keywords and
treat all Modelica names equal instead.

In summary we have:
```
  name_ : Modelica name in functions
  name  : local variable of Cpp runtime
 _name  : member variable of Cpp runtime
 _name_ : Modelica name as member variable/define for equations
```


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-01-30_17-41-46](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3829/console).

PR #1440, opened 2017-01-30
Add support for firstTick to c runtime

lochel:
I am not really sure if we want to support `firstTick` with the c runtime right now.


lochel:
@adrpo @rfranke Right now `firstTick` is treated as Modelica build-in call in the front end and cpp runtime. Do we want it that way, because it is not yet part of the specification? To be more Modelica compliant, we could introduce a new flag to extended the specification by any MCP, e.g. `--mcp=24`. What do you think?

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-02-16_14-33-26](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3916/console).

PR #1445, opened 2017-01-31
NFInst prefixing improvements.

perost:
- Changed Lookup.lookupCref to return a list of all found nodes instead
  of a prefix, so the prefix can be properly constructed by the caller.
- Fixed prefixing in Typing.typeCref, which previously typed the prefix
  incorrectly.
- Moved typing of builtin attributes from typeComponent to typeBinding,
  to avoid infinite loops with enumerations.
- Fixed handling of builtin functions size and smooth to use the correct
  arguments.

### DIFF
--- a/Compiler/NFFrontEnd/NFFunc.mo
+++ b/Compiler/NFFrontEnd/NFFunc.mo
@@ -89,7 +89,7 @@ protected
   String fn_name;
   Absyn.Path fn, fn_1;
   InstNode fakeComponent;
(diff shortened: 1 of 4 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-01-31_15-09-26](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3841/console).

PR #1460, opened 2017-02-08
Handle double matrix when strings are expected

sjoelund:
This handles some versions of matlab that save character or integer
matrices using double precision.


sjoelund:
@niklwors If this works, you can read your Matlab-generated files in OMC. I would suggest fixing the export though. Matlab deciding to store the character 'A' as 65.0 in double precision just seems odd (and takes 8x the storage space).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-02-08_09-49-42](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3870/console).

PR #1471, opened 2017-02-10
Test arrayGet reporting errors for out of bounds

sjoelund:
This is so @adrpo can investigate the sources of array out of bounds.
It is better if we do not get these since they tend to not show errors
for the user.


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-02-10_09-34-52](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3889/).

sjoelund:
@adrpo: Is this still needed?

PR #1494, opened 2017-02-20
Fix extends modifier issue in NFInst.

perost:
- Fixed extends modifier issue by cloning base classes, so that
  modifiers aren't applied to the class tree.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-02-20_10-59-59](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3930/console).

PR #1496, opened 2017-02-20
Frontend: Do not evaluate external object inputs

sjoelund:
This fixes ticket:4207, where calling the external object constructor in
the correct order was fixed. It is then safe to not force evaluation of
its inputs.

### DIFF
--- a/Compiler/FrontEnd/Static.mo
+++ b/Compiler/FrontEnd/Static.mo
@@ -612,7 +612,7 @@ algorithm
     (outCache, {tty}) := Lookup.lookupFunctionsInEnv(inCache, inEnv, path, inInfo);
     tty := Types.makeFunctionPolymorphicReference(tty);
     (outCache, args, consts, _, tty, _, slots) := elabTypes(outCache, inEnv, pos_args,
-      named_args, {tty}, true, true, inImplicit, NOT_EXTERNAL_OBJECT_MODEL_SCOPE(),
+      named_args, {tty}, true, true, inImplicit,
       NONE(), inPrefix, inInfo);
     if not Types.isFunctionPointer(tty) then
       (outCache, path) := Inst.makeFullyQualified(outCache, inEnv, path);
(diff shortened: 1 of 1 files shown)


sjoelund:
@lochel - do you now test that all inputs of the external object have been evaluated before calling the constructor in the backend? If so, this should be safe to merge.

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-02-20_13-49-39](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3931/).

lochel:
Yes, I think that should be checked with #1341.

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-02-20_14-06-02](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3932/).

sjoelund:
It seems we now allow things like:
```mo
model ExternalObjectVariableModel
  Integer a = integer(c);
  Real c(fixed=false, start=0);
  parameter Integer b = 2;
  ExtObj.Object obj = ExtObj.Object(a);
initial equation
  c = der(c);
equation
  der(c) = 0.1;
end ExternalObjectVariableModel;
```

I am not sure that is a good idea... (Just taking the value at the initial point instead of making sure it is a parameter expression). Although, it seems the specification doesn't prohibit it.

sjoelund:
@bernhard-thiele @adrpo Does https://gist.github.com/sjoelund/cca64be26d79a45e6f96d226a3c8e043 work in Dymola?

bernhard-thiele:
What do you mean with "work"? The gist references `#include \"mainfunctions.h\"", Library = {"libexternalObject.a"}` which is of course not there and results in an error during compilation of the C sources. But it translates the model to C sources.

sjoelund:
I guess I can update the expected output then. Remove some tests...

lochel:
Does `ExternalObjectVariableModel` compile with OpenModelica?

sjoelund:
Yes...

lochel:
Okay, I will check that. I thought that I had already implemented a check for that ...

PR #1498, opened 2017-02-21
Check if all inputs to external object constructors have been evaluated before calling them

lochel:
This is related to #1341 and #1496.

### DIFF
--- a/Compiler/BackEnd/Initialization.mo
+++ b/Compiler/BackEnd/Initialization.mo
@@ -797,6 +797,8 @@ protected
   array<Integer> secondary;
   BackendDAE.Var p;
   DAE.Exp bindExp;
+  HashSet.HashSet hs;
+  list<DAE.ComponentRef> crefs;
 algorithm
   outInitVars := selectInitializationVariables(dae.eqs);
   outInitVars := BackendVariable.traverseBackendDAEVars(dae.shared.globalKnownVars, selectInitializationVariables2, outInitVars);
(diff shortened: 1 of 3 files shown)


lochel:
@sjoelund One need to test if all inputs to external object constructors are of parameter variability. I can do it in the backend, but maybe it would be better to check it in the frontend. What do you think?

sjoelund:
You can't really check it in the frontend. External objects are allowed to be passed non-parameters. I'd say it's discouraged to pass it a time-varying signal though...

lochel:
> External objects are allowed to be passed non-parameters.

Well, that is currently not supported by OpenModelica…

sjoelund:
No, it is supported :) In functions, we can construct external objects just fine.

lochel:
But not yet within models …

sjoelund:
So it makes sense to check it in the backend, then. And use Error.mo message for the error so gettext is used and the message can be translated :)

lochel:
I will not add error messages for that, but support it properly in the back end. It seems that there are some good reasons to do it, e.g. it is used within the DeviceDrivers library.

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-02-21_14-53-09](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3942/).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-02-21_15-51-15](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3943/console).

PR #1501, opened 2017-02-21
Fix KinsolLapack.cpp with sundials 2.7.0

jschueller:
This fixes this compilation error in KinsolLapack.cpp:
```
/openmodelica-git/src/openmodelica-git/OMCompiler/SimulationRuntime/cpp/Solver/Kinsol/KinsolLapack.cpp: In function ‘int KINLapackCompletePivoting(void*, int)’:
/openmodelica-git/src/openmodelica-git/OMCompiler/SimulationRuntime/cpp/Solver/Kinsol/KinsolLapack.cpp:44:24: error: invalid conversion from ‘void (*)(KINMem) {aka void (*)(KINMemRec*)}’ to ‘int (*)(KINMemRec*)’ [-fpermissive]
kin_mem->kin_lfree = KINLapackCompletePivotingFree;
```

### DIFF
--- a/SimulationRuntime/cpp/Include/Solver/Kinsol/KinsolLapack.h
+++ b/SimulationRuntime/cpp/Include/Solver/Kinsol/KinsolLapack.h
@@ -12,7 +12,7 @@ static int KINLapackCompletePivotingSolve(KINMem kin_mem, N_Vector x, N_Vector b
 #else
 static int KINLapackCompletePivotingSolve(KINMem kin_mem, N_Vector x, N_Vector b, realtype *res_norm);
 #endif
-static void KINLapackCompletePivotingFree(KINMem kin_mem);
+static int KINLapackCompletePivotingFree(KINMem kin_mem);
 static int calcJacobian(KINMem kin_mem);
 
 struct linSysData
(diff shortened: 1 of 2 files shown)


OpenModelica-Hudson:
The job failed to compile; for details, see [OpenModelica_TEST_PULL_REQUEST 2017-02-21_22-56-24](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3944/console).

sjoelund:
You will need to #ifdef your patch for the sundials version that caused these changes.

jschueller:
Right, this has changed in sundials latest 2.7.0.
Sundials provides SUNDIALS_PACKAGE_VERSION macro but it's not very usable as it's a string:
```sundials_config.h:18:#define SUNDIALS_PACKAGE_VERSION "2.7.0"```

sjoelund:
I was going to say in that case, you would need to detect the API in autoconf, but that project uses cmake so that's going to be really annoying for whoever fixes this :)

⁣Sent from BlueMail ​

niklwors:
we already check the sundials version in the  CMake  file SimulationRuntime/cpp/CMakeLists.txt.  So normaly it should compile with all sundials versions.

jschueller:
@niklwors I don't see a different code path for sundials 2.7

niklwors:
@jschueller Do you mean in KinsolLapack.cpp? You can add an additional definition
 #if (SUNDIALS_MAJOR_VERSION == 2 && SUNDIALS_MINOR_VERSION >7)
I

niklwors:
this should be fixed in #1519

PR #1505, opened 2017-02-24
fixed division by zero error and added homotopy solver

qichenghua:
fixed issue that occured if evaluating the algloop at the initial guess throws a division by zero error and added homotopy solver

### DIFF
--- a/SimulationRuntime/cpp/Include/Solver/Nox/Nox.h
+++ b/SimulationRuntime/cpp/Include/Solver/Nox/Nox.h
@@ -30,6 +30,9 @@ private:
   void solverinit();
   void createStatusTests();
   void createSolverParameters();
+  void LocaHomotopySolve(int numberofhomotopytries);
+  NOX::StatusTest::StatusType BasicNLSsolve();
+  NOX::StatusTest::StatusType secondBasicNLSsolve();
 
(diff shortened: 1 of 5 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-02-24_10-02-32](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/3951/console).

sjoelund:
@qichenghua Since no test changed, I think you should add a testcase or two to the test suite that uses the new code.

qichenghua:
@sjoelund The fix for the initial guess issue, which occured in the three springs model, was implemented in the Nox solver. We are trying to get the Trilinos package to work, which the nonlinear solver Nox depends on. Once that is done, the nonlinear solver Nox can be tested with the entire test suite. The modification in Kinsol has only changed the error message printed for example in the three springs model.

lochel:
So does the cpp runtime now support the Modelica `homotopy` operator for initialization?

qichenghua:
Unfortunately, not. The nonlinear solver Nox now supports homotopy methods for solving systems of nonlinear equations similar to those used in the c runtime.

PR #1539, opened 2017-03-12
NFFrontEnd function updates.

mahge:
- Instantiate overloaded functions one by one and propagate multiple overlaods.
    - OVERLOADED_CLASS is not used for overloaded functions anymore. We keep things in the cache.
    - Do not overwrite function names by the overload name.
    - Make sure we can't overlaod no function classes...
  - Do not match args while insting them.
    - We need the positional and named args separtely until we reach typing.
    - If we match and try to replace named args by values then overlaoding will not work properly.
  - Create slots after typing a function not when insting. We need the type info in the slots.
  - Added a new Call type ARG_TYPED_CALL
    - this is used while typing a call t preserve type and variablity info.
  - Track errors and print if no match only.
  - Save type info when filling args from slot defaults.
  - Handle printing of arr_typed_call.
  - Some more changes...

### DIFF
--- a/Compiler/NFFrontEnd/NFCall.mo
+++ b/Compiler/NFFrontEnd/NFCall.mo
@@ -48,6 +48,8 @@ import Typing = NFTyping;
 import Types;
 import List;
 import NFClass.Class;
+import ErrorExt;
+import Util;
 
 public
 uniontype CallAttributes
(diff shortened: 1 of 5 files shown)


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-03-12_15-03-59](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4013/).

PR #1541, opened 2017-03-13
Simplify of asub where the subscript is a range

sjoelund:
This handles some obscure code that gets inlined. Ranges are not
simplified in the general case because they can be huge and usually are
used to iterate over some indexes without the need to construct an
array. Subscripts are different in that they are used to construct
arrays, so they should be safe to evaluate at compile-time.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-03-13_15-01-35](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4018/console).

PR #1550, opened 2017-03-15
Nf function improvments 2

mahge:
- Instantiate overloaded functions one by one and propagate multiple overloads.
    - OVERLOADED_CLASS is not used for overloaded functions anymore. We keep things in the cache.
    - Do not overwrite function names by the overload name.
    - Make sure we can't overload non function classes...
  - Do not match args while insting them.
    - We need the positional and named args separately until we reach typing.
    - If we match and try to replace named args by values then overloading will not work properly.
  - Create slots after typing a function not when insting. We need the type info in the slots.
  - Added a new Call type ARG_TYPED_CALL
    - this is used while typing a call t preserve type and variability info.
  - Track errors and print if no match only.
  - Save type info when filling args from slot defaults.
  - Handle printing of arr_typed_call.
  - Some more changes...

 NFFrontEnd function and typing improvements

  - Overloading is now resolved with priority to exact match if there is any.
  - Overwrite overload resolved name back to the overload name for builtin functions
    - if OpenModelica.Internal.intAbs is matched we just rename the call to abs().
    - Only for builtin functions because they are not collected anyway.
  - If overloading fails print the overload name instead of the actual names in the errors.
  - TypeMatch now notifies how matching is done (exact, cast, generic or UNKNOWN)
  - Function matching now notifies how it is matched (exact, cast, generic or vectorized).

### DIFF
--- a/Compiler/NFFrontEnd/NFCall.mo
+++ b/Compiler/NFFrontEnd/NFCall.mo
@@ -48,6 +48,8 @@ import Typing = NFTyping;
 import Types;
 import List;
 import NFClass.Class;
+import ErrorExt;
+import Util;
 
 public
 uniontype CallAttributes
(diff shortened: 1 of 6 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-03-16_10-00-47](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4041/console).

PR #1556, opened 2017-03-17
Check if binding type does match the component type

lochel:
This introduces a check for the component binding type. It should match the component type (or be implicitly convertible).

@perost Is this the intended way of implementing it?
@mahge It seems that there is a problem with overloaded functions. Can you please have look at FuncOverloadMulti.mo? It generates now the following error message:
```
[NFTyping.mo:307:13-307:164:writable]Modelica Assert: NFTyping.typeComponentBinding got type mismatch between component x and its binding bool_string(true)!
Error processing file: FuncOverloadMulti.mo
```

### DIFF
--- a/Compiler/NFFrontEnd/NFBinding.mo
+++ b/Compiler/NFFrontEnd/NFBinding.mo
@@ -121,6 +121,13 @@ public
     end match;
   end typedExp;
 
+  function getTypedExp
+    input Binding binding;
+    output Expression exp;
+  algorithm
+    TYPED_BINDING(bindingExp = exp) := binding;
+  end getTypedExp;
+
   function setTypedExp
     input Expression exp;
     input output Binding binding;
@@ -174,6 +181,16 @@ public
       else false;
     end match;
   end isEach;
+
+  function isTyped
+    input Binding binding;
+    output Boolean isTyped;
+  algorithm
+    isTyped := match binding
+      case TYPED_BINDING() then true;
+      else false;
+    end match;
+  end isTyped;
 
   function toString
     input Binding binding;
--- a/Compiler/NFFrontEnd/NFTyping.mo
+++ b/Compiler/NFFrontEnd/NFTyping.mo
@@ -51,19 +51,20 @@ import NFType.Type;
 import Operator = NFOperator;
 
 protected
-import Inst = NFInst;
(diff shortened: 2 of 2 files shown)


lochel:
Well, I have to change the assert to a proper error message...

mahge:
The error message on FuncOverloadMulti.mo is also correct. The functions real_string and bool_string should return String not Integer. My bad. You can update them or I will.

Compiler/NFFrontEnd/NFTypeCheck.mo
```diff
@@ -1448,8 +1448,8 @@ function matchTypes
   input Type expectedType;
   input output Expression expression;
   input Boolean allowUnknown = false;
-        output Type compatibleType;
-        output Boolean compatible = true;
+  output Type compatibleType;
```
perost: I indent outputs in this way on purpose when using input output, since it makes it easier to spot all the outputs (notice that these two outputs line up with 'expression' above). So please don't change this.

perost:
>@perost Is this the intended way of implementing it?
Yeah, that seems about right. We need to handle propagated dimensions too, but that's something I'm working on so just ignore that for now.

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-03-17_14-42-31](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4048/).

Compiler/NFFrontEnd/NFTypeCheck.mo
lochel: I see! I will restore the whitespaces.

However, I think it would be better to emphasize the outputs using the following formatting:
```
function matchTypes
  input        Type actualType;
  input        Type expectedType;
  input output Expression expression;
  input        Boolean allowUnknown = false;
        output Type compatibleType;
        output Boolean compatible = true;
protected
…
```

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-03-26_17-36-22](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4093/console).

PR #1560, opened 2017-03-20
NFInst improvements.

perost:
- Added Expression.TYPENAME to handle typenames better.
- Use RangeIterator more to simplify scalarization.
- Improved handling of for loops.
- Improved iteration ranges.
- Cleaned up type checking:
  - Removed erroneous cases for Type.FUNCTION.
  - Changed MatchKind into an enumeration.
  - Removed boolean return value from type matching functions,
    not needed when we also return MatchKind.

### DIFF
--- a/Compiler/NFFrontEnd/NFBinding.mo
+++ b/Compiler/NFFrontEnd/NFBinding.mo
@@ -121,6 +121,19 @@ public
     end match;
   end typedExp;
 
+  function setTypedExp
+    input Expression exp;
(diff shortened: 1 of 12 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-03-20_12-10-38](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4054/console).

PR #1561, opened 2017-03-20
Improve module wrapFunctionCalls

ptaeuber:
See [ticket:3812](https://trac.openmodelica.org/OpenModelica/ticket/3812)

- constant cse-variables are stored in globalKnownVars and **only
  evaluated once** before simulation instead of in each simulation step
- also traverse globalKnownVars for function calls before
  traversing the systems
- SimEqns for parameters are created from the globalKnownVars
- constant variables (eg. cse-variables) are solved before initialization in
  updateBoundParameters in *_08bnd.c
- primaryParameters and allPrimaryParameters are removed
- improved dumps and documentation

### DIFF
--- a/Compiler/BackEnd/CommonSubExpression.mo
+++ b/Compiler/BackEnd/CommonSubExpression.mo
@@ -32,8 +32,8 @@
 encapsulated package CommonSubExpression
 " file:        CommonSubExpression.mo
   package:     CommonSubExpression
-  description: This package contains functions for the optimization module
-               CommonSubExpression."
+  description: This package contains functions for the optimization modules
+               wrapFunctionCalls, commonSubExpressionReplacement and cseBinary."
 
 
 public
@@ -56,6 +56,7 @@ import ExpressionDump;
 import ExpressionSolve;
 import ExpressionSimplify;
 import Global;
+import HashSet;
 import HashTableExpToExp;
 import HashTableExpToIndex;
 import HpcOmTaskGraph;
@@ -73,6 +74,8 @@ end CSE_Equation;
 
 constant CSE_Equation dummy_equation = CSE_EQUATION(DAE.RCONST(0.0), DAE.RCONST(0.0), {});
 constant Boolean debug = false;
+constant String BORDER = "###############################################################";
+constant String UNDERLINE = "========================================";
 
 protected function printCSEEquation
   input CSE_Equation cseEquation;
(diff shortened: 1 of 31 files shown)


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-03-20_14-00-26](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4056/).

ptaeuber:
@adrpo As far as I can tell [it](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4056/testReport/junit/(root)/openmodelica_cppruntime_genDebugSymbols/Modelica_Fluid_Examples_BranchingDynamicPipes_genDebugSymbols_mos/) is just a simple warning for consistency of start value and actual value for constant variables. It did not appear before because constant variables were not evaluated together with the parameters. So I will update the output.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-03-21_14-25-04](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4067/console).

casella:
@ptaeuber did you check that the execution time scales at worst as O(N log(N)) with your latest additions?

ptaeuber:
@casella hm no, I will do some testing

casella:
Thanks! It is important that we avoid adding methods that do smart things but scale badly because of not careful enough implementation

sjoelund:
The thing I'm mostly worried about is the copying of hashtables. If the number of independent equation systems is large (which is not often the case), that is an insane bottleneck. Perhaps try using a trie or AvlTree to do lookups instead...

lochel:
Actually, the hash tables are not needed at all.

adrpo:
@ptaeuber: It seems that many libraries took a dive in regression testing:
https://test.openmodelica.org/libraries/trend-detailed.html

For example:
https://test.openmodelica.org/libraries/ThermoSysPro/BuildModelRecursive.html
https://test.openmodelica.org/libraries/ThermoSysPro/files/ThermoSysPro.Examples.SimpleExamples.TestStodolaTurbine1.err
Error: Too many equations, over-determined system. The model has 144 equation(s) and 143 variable(s).
Error: Internal error Transformation Module PFPlusExt index Reduction Method Pantelides failed!
Notification: Performance of <failed> postOpt wrapFunctionCalls (simulation): time 0.004559/6.552, allocations: 0.8966 MB / 1.188 GB, free: 105.8 MB / 462.1 MB
Error: post-optimization module wrapFunctionCalls (simulation) failed.

ptaeuber:
@adrpo That's bad. I will revert the commit for now until everything is fine.
@ lochel Why do you think the hashtable is not needed? The whole algorithm is based on that hastable...

Compiler/BackEnd/CommonSubExpression.mo
```diff
@@ -94,105 +97,199 @@ algorithm
...
   for syst in inDAE.eqs loop
-    HT := BaseHashTable.clear(HT);
-    exarray := ExpandableArray.clear(exarray);
+    if Flags.isSet(Flags.DUMP_CSE_VERBOSE) then
+      print("\n\nHandle system (belongs to " + daeTypeStr + " DAE):\n" + BORDER + "\n");
+      BackendDump.dumpVariables(syst.orderedVars, "Variables");
+      BackendDump.dumpEquationArray(syst.orderedEqs, "Equations");
+      print("\n\nPhase 1: Analysis\n" + BORDER + "\n");
+    end if;
+    HT := BaseHashTable.copy(HT_initial);
```
ptaeuber: @sjoelund I think you mean this copying? I did not know that it is such a bottleneck... I will probably remove the HT_initial then

lochel:
@ptaeuber Well, I guess you are right. Just ignore my comment 😉

Compiler/BackEnd/CommonSubExpression.mo
sjoelund: Yes,  a hashtable is always expensive to create because you typically create an array of size relative to all equations: O(n). And here you copy it for each system of equations (also O(n), though often small). O(n^2) is bad though, and if copying lookup tables is necessary, an immutable structure like a trie or tree which has a constant-time copy (just a pointer copy) is preferred.

PR #1563, opened 2017-03-20
fix codegeneration for records

wibraun:


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-03-20_18-15-04](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4060/console).

lochel:
@wibraun Unfortunately, you did not provide a description for this pull request. So what did you fix with it?

The changes did also not affect the test suite, so probably you should add a test case for this.

wibraun:
@lochel Unfortunately, it's sometimes quite hard to setup specific test-case, and have only a really big model where this occurs. But if you look in the code it's obviously wrong to generate equations, which does not check if the lhs expression is really the desired expression.

casella:
I provided a test case which is proprietary, so it cannot be added to the testsuite

PR #1572, opened 2017-03-23
compute time events during runtime instead of precalculation (cpp)

vwaurich:
changed handling of time events in cpp runtime
now we have:
- initTimeEventData() to initialize the time event definition from the model (starttime, interval)
- computeNextTimeEvent() to compute the time for the upcoming time event
- computeTimeEventConditions() to compute the conditions for the time events at a certain point in time
- resetTimeEvents

This is also implemented in the FMUWrapper 1 and 2. Time events havent been available so far for FMI.
Discrete Time FMUs are not yet tested.
Hudson is fine with that except clockedSolverTest.mos needs reference update (we now write the right limit of a time event step instead the left limit as before, now its like in Dymola)


vwaurich:
@rfranke @niklwors  You might have a look.

lochel:
@vwaurich What kind of events are handled as time events?

vwaurich:
@lochel time events resulting from sample() and clock-events

PR #1573, opened 2017-03-23
Fix initialization of variable attributes

lochel:
see [ticket:4359](https://trac.openmodelica.org/OpenModelica/ticket/4359)

---

**TODO**
- [x] Fix code generation for c runtime
- [x] Fix code generation for cpp runtime
- [x] Fix openmodelica/fmi/ModelExchange/1.0/JuliansBib.mos
- [x] Fix openmodelica/xml/Bug3857.mos
- [x] Do some more testing

PR #1592, opened 2017-04-04
Fix initialization of min/max/nominal attributes

lochel:
This is about fixing initialization of attributes that depend on secondary parameters. Start attributes were already fixed with #1573.

---

**TODO**

- [ ] Fix min attributes
- [ ] Fix max attributes
- [ ] Fix nominal attributes
- [ ] Testing


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-04-04_15-42-02](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4144/console).

lochel:
I think we should throw a warning if min, max, or nominal attributes depend on secondary parameters (those having fixed=false) and then just ignore them.

PR #1601, opened 2017-04-07
Remove $_old(.)

lochel:
`$_old(.)` is not handled in the back end properly and may introduce loops. It seems that it is also not necessary, and pre values can be used instead.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-04-07_10-27-41](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4165/console).

PR #1619, opened 2017-04-15
Improved performance and reduced memory consumption

lochel:


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-04-16_00-12-33](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4207/).

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-04-16_23-31-25](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4209/).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-04-17_11-42-55](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4210/console).

casella:
@lochel does this PR have any impact on the issues reported in [#4146](https://trac.openmodelica.org/OpenModelica/ticket/4146), as far as you know?

lochel:
This should have some impact on the general performance and memory consumption. However, I guess the impact is not that big.
I will create some statistics that cover all the different phases and show the development over the last few weeks to see if we have any improvements. Maybe we can run then this regularly together with the coverage test.

casella:
I think (and I am not alone in this, see, e.g. [this page](https://en.wikipedia.org/wiki/Program_optimization#When_to_optimize)) that optimizations should really be test-driven. 

There are so many places where the code is (in principle) inefficient that fixing them all is not possible, so we should focus on those chunks of code which really hurt in some relevant use case. It doesn't matter where the use case comes from, but there should be one. Was there one for this specific PR?

In general, I would recommend that all PRs related to code optimizations refer to some tests that give an objective measurement of the achieved performance improvement. Otherwise, the risk of wasting precious developers' time for marginal gains is very high.

I have opened [#4146](https://trac.openmodelica.org/OpenModelica/ticket/4146) to collect those bottlenecks that I have encountered playing around with large grid models and the ScalableTestSuite. In 5 months, only 1 out of 6 issues has been addressed and resolved. I would invite people to add more relevant test cases and refer to them there, using this ticket as a collecting point for interesting and relevant backend scaling issues that are worth spending time and effort on.

The most interesting fixes would be those ones that improve the performance of frequently used inefficient functions, because they have the maximum likelyhood of having a broad impact on performance. However, spotting these improvements out proves especially hard to do in practice. For example @perost spotted an inefficient O(N) instead of O(1) implementation of addEquation, which is used all over the place, and rewrote it, but apparently that had no significant impact on performance on the ScalableTestSuite.

My 2 cts :)

lochel:
This was mostly about removing redundant code and removing side-effects that are the cause of nasty bugs. The optimizations that I collected with the PR are just a by-product.
Before, we allocated always 40% extra space for equation arrays - just in case. This is one of the things that I've removed.

PR #1621, opened 2017-04-18
Reimplementation of homotopy approach

lochel:
**TODO**

- [X] Remove old approach
- [x] Add homotopy flag to nonlinear systems
- [X] Extend nonlinear solvers to basic homotopy support
- [x] Remove BackendDAE and code generation for lambda=0
- [X] Improve logging information
- [ ] More testing

### DIFF
--- a/Compiler/BackEnd/BackendDAEUtil.mo
+++ b/Compiler/BackEnd/BackendDAEUtil.mo
@@ -6818,8 +6818,6 @@ public function getSolvedSystem "Run the equation system pipeline."
   output BackendDAE.BackendDAE outSimDAE;
   output BackendDAE.BackendDAE outInitDAE;
   output Option<BackendDAE.InlineData > outInlineData;
-  output Boolean outUseHomotopy "true if homotopy(...) is used during initialization";
-  output Option<BackendDAE.BackendDAE> outInitDAE_lambda0;
   output list<BackendDAE.Equation> outRemovedInitialEquationLst;
 protected
   BackendDAE.BackendDAE dae, simDAE;
@@ -6877,7 +6875,7 @@ algorithm
   end if;
 
   // generate system for initialization
-  (outInitDAE, outUseHomotopy, outInitDAE_lambda0, outRemovedInitialEquationLst, globalKnownVars) := Initialization.solveInitialSystem(dae);
+  (outInitDAE, outRemovedInitialEquationLst, globalKnownVars) := Initialization.solveInitialSystem(dae);
 
   // use function tree from initDAE further for simDAE
   simDAE := BackendDAEUtil.setFunctionTree(dae, BackendDAEUtil.getFunctions(outInitDAE.shared));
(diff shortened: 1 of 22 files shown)


lochel:
It seems that parts of this are already implemented with ecd3174aaa5088b65058093b0598b0c7410d733a.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-04-19_13-35-27](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4214/console).

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-04-19_18-26-21](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4216/).

lochel:
@casella The number is arbitrary and only used for testing. The number of steps can be set with `-ils=<number>`. It's not decided yet if we will enable homotopy with this PR or not.

casella:
OK. So you might disable homotopy by default before pushing it to master after testing?

lochel:
Yes, exactly. Do you have any opinion on this?

It would be interesting to see the impact on the coverage test at least. But maybe we should not enable it by default before using a proper homotopy solver.

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-04-20_12-45-26](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4218/).

casella:
Dymola's approach is (or at least was, some time ago) to always try to initialize without homotopy, and then only use homotopy as a fallback. The pro argument of doing this is that going with homotopy right away possibly wastes time, if it's not needed to converge. The con argument is that one may use homotopy to steer the solver to the correct solution, so not using it may lead to convergence to the wrong one.
All things considered, if the coverage is not worsened (or possibly improves) I think the con argument is stronger, so I would turn it on by default even with the basic implementation we have now. If we considere the whole compile+initialize+simulate time, the overhead of running a few steps of crude homtopy is marginal, while improved convergence of initialization is a big deal. Note that homotopy is now used in the MSL, even in basic components such as the saturation block. Of course we need to make sure that the crude approach gives some real benefit on the testsuite, but if it does, why not exploiting it?

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-04-20_13-00-59](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4219/).

lochel:
I will push this without activating homotopy and I will create a new PR just to activate it so that we can easily revert it if necessary.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-04-20_13-46-52](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4220/console).

casella:
OK, excellent

PR #1623, opened 2017-04-20
Enable homotopy by default

lochel:
**TODO**

- [ ] Some more testing

### DIFF
--- a/SimulationRuntime/c/simulation/solver/nonlinearSystem.c
+++ b/SimulationRuntime/c/simulation/solver/nonlinearSystem.c
@@ -53,7 +53,7 @@
 
 int check_nonlinear_solution(DATA *data, int printFailingSystems, int sysNumber);
 
-int init_lambda_steps = 1;
+int init_lambda_steps = 4;
 
 struct dataNewtonAndHybrid
 {
--- a/SimulationRuntime/c/util/simulation_options.c
+++ b/SimulationRuntime/c/util/simulation_options.c
@@ -142,7 +142,7 @@ const char *FLAG_DESC[FLAG_MAX+1] = {
   /* FLAG_IIF */                   "value specifies an external file for the initialization of the model",
   /* FLAG_IIM */                   "value specifies the initialization method",
   /* FLAG_IIT */                   "[double] value specifies a time for the initialization of the model",
-  /* FLAG_ILS */                   "[int] default: 1",
+  /* FLAG_ILS */                   "[int] default: 4",
   /* FLAG_IMPRK_ORDER */           "[int (default 5)] value specifies the integration order of the implicit Runge-Kutta method. Valid values: 1-6",
   /* FLAG_IMPRK_LS */              "selects the linear solver of the integration methods: impeuler, trapezoid and imprungekuta",
   /* FLAG_INITIAL_STEP_SIZE */     "value specifies an initial step size for supported solver",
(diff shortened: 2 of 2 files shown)


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-04-23_12-21-11](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4225/).

lochel:
I will just push this to see the impact on the coverage tests.

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-04-26_13-53-06](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4231/).

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-04-26_14-16-51](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4232/).

casella:
Sounds good! I will now start a test of ThermoPower_Experimental, just in case

lochel:
This is the impact on the coverage tests: [2017-04-26 05:50:18..2017-04-27 06:03:35.html](http://libraries.openmodelica.org/branches/history/master/2017-04-26%2005:50:18..2017-04-27%2006:03:35.html)

casella:
I gave a quick look at the impact on coverage tests. On one hand, there is a positive effect on the initialization of may fluid models, which is expected. On the other hand, I see a very drastic increase of simulation times for many models (fluid and not fluid). This is strange: usually the time spent for initialization is a few percent of the overall simulation time, so multiplying it by 5 or 10 (what was exactly the argument of -ils?) should not change the simulation time by one order of magnitude.
Did you make some analysis of where is all this extra time spent and why?

lochel:
The test was done using –ils=4 as default choise. However, I’ve reverted the changes because the results are unexpectedly bad. The plan is now as follows: 
* Investigation of the regressions, both simulation and performance. Then we will see if there is a critical bug in the current implementation.
* Utilizing a special homotopy solver to follow properly the homotopy path.
* Providing a symbolically simplified system for lambda=0.

ptaeuber:
@casella @lochel I had a look into some models which simulate a lot slower now.
The time is lost in the calculation of the lambda=0 system because of function range asserts.
E.g. in `ThermoPower.Examples.CISE.Simulators.CISESim180503`:
`assert            | error   | Error in region computation of IF97 steam tables(p = 3.41346e+06, h = -1.61499e+07)`
or
`assert            | error   | IF97 medium function visc_dTp: viscosity computed outside the range
|                 | |       | of validity of the IF97 formulation: p = 3.40694e+06 Pa, Tc = 912.774 K`

The same applies for example for `Modelica.Fluid.Examples.HeatExchanger.HeatExchangerSimulation`. Here the system even fails if the homotopy call is replaced by the simplified part in the Frontend.

So, for me this sounds like a modeling problem rather than a tool issue.
@casella What do you think?

casella:
@lochel: I understood that the simplified system for lambda = 0 had been in place for a while. Was I mistaken?

@ptaeuber I'm really pressed with the Modelica Conference Proceedings now. I'll look into that when I'm finished with them

lochel:
A new try: #1748

PR #1630, opened 2017-04-24
fix ScalableTestSuite.Thermal.Advection.ScaledExperiments.SteamPipe

hkiel:
use nominal attribute only on Real variables
fix parameter type

### DIFF
--- a/Compiler/Template/CodegenC.tpl
+++ b/Compiler/Template/CodegenC.tpl
@@ -2837,7 +2837,7 @@ template getIterationVars(list<ComponentRef> crefs, String indexName)
   ;separator="\n")
   <<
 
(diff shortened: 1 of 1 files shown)


hkiel:
Can you please restart ScalabeTestSuite once this is PR merged?

hkiel:
@wibraun Can you please check if the change is okay, because now `algebraicNominal[]` contains fewer element than before (non-Reals are missing now). Maybe the size of the array must be adjusted somewhere, or should it be filled with zero values (maybe it is already done by some calloc somewhere)?

wibraun:
@hkiel As far as I see this needs a fix in the Backend and can't be fixed at code generation level, since we can't pass integer variables to the dae solver, it expects that all variables are continuous.

PR #1631, opened 2017-04-24
move simple functions to static inline in model_help.[ch]

hkiel:
avoids unnecessary function calls and could help compiler to optimize relations

### DIFF
--- a/SimulationRuntime/c/simulation/solver/model_help.c
+++ b/SimulationRuntime/c/simulation/solver/model_help.c
@@ -1256,43 +1256,12 @@ modelica_boolean LessZC(double a, double b, modelica_boolean direction)
   return direction ? (a - b <= eps) : (a - b <= -eps);
 }
 
-modelica_boolean LessEqZC(double a, double b, modelica_boolean direction)
-{
-  return !GreaterZC(a, b, !direction);
-}
-
 /* TODO: fix this */
 modelica_boolean GreaterZC(double a, double b, modelica_boolean direction)
 {
   double eps = tolZC * fmax(fabs(a), fabs(b)) + tolZC;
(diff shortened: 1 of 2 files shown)


hkiel:
Can you please restart ScalabeTestSuite once this is PR merged?

sjoelund:
I guess this should be benchmarked since we compile code with -O0 by default. Or maybe we should change it to -O1 or -Og so inlining, etc works as expected...

hkiel:
@sjoelund Where is the default optimization level set? I cannot find "-O0" in *.mo files.

hkiel:
@sjoelund @casella
Ok, I use `setCFlags("-O1")` now and "-n=1" for omc.
I simulated `ScalableTestSuite.Electrical.BreakerCircuits.ScaledExperiments.BreakerNetworkDelayed_N_320` and  compile time went up from 10s -> 37s, while simulation time went down from 65s -> 45s.
When using "-Os" compile time eveng goes up to 55s and simulation time down to 34s.

An impact of this modification could not be seen.

I tested only one model, but overall I guess we have a slower system when C-compiler is optimizing.
Maybe users can switch on optimization for really slow simulations (there are some in the ScalableTestSuite)...

sjoelund:
Yes, users can do turn on the optimiser flags. But in general in most people don't, even when benchmarking OpenModelica simulations...

hkiel:
Can we emit a hint to the user that turning on optimization can speed up things?
Maybe that can be shown when simulation takes more than a few seconds.

hkiel:
I'm closing this, because C-compiler does not take advantage of this when not optimizing, and when optimizing, compilation slow down is larger than speed up of simulation run.

PR #1634, opened 2017-04-27
Parse mos-file statements using iteration

sjoelund:
Recursion caused very long mos-files to crash during the parsing. Doing
a listReverseInPlace is a small price to pay for not crashing.
Note: There will be no testcase for this since you need ~45MB mos-files
to trigger this on an -Os optimized omc (really only triggers for
-O0 -g).

### DIFF
--- a/Parser/Modelica.g
+++ b/Parser/Modelica.g
@@ -1676,19 +1676,23 @@ interactive_stmt returns [void* ast]
 @declarations { int last_sc = 0; }
 @init{ ss = 0; } :
   // A list of expressions or algorithms separated by semicolons and optionally ending with a semicolon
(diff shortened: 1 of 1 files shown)


lochel:
Why not providing a small test case that generates a large mos-file as actual test case?

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-04-27_13-17-14](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4238/console).

sjoelund:
Because it takes a lot of time to parse and execute 3 500 000 statements (6m49.938s for something as simple as `1;`, 3566556 times).

PR #1638, opened 2017-05-02
Fix ticket4407 case preserving filesystems

hkiel:
fixes problem with re-generating files on case-preserving filesystems
- removes files (with wrong case) before creating with correct case again

### DIFF
--- a/Compiler/Util/omc_file.h
+++ b/Compiler/Util/omc_file.h
@@ -100,7 +100,16 @@ static inline void om_file_open(__OMC_FILE *file, const char *filename, int mode
 #endif
     fclose(file->file);
   }
+#if defined(__APPLE_CC__)||defined(__MINGW32__)||defined(__MINGW64__)
+  if (mode == 1) {
+    file->file = fopen(filename, "rb");
+  } else {
+    unlink(filename);
+    file->file = fopen(filename, "wb");
+  }
+#else
   file->file = fopen(filename, mode == 1 ? "rb" : "wb");
+#endif
   file->name = filename;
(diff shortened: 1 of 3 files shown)


hkiel:
@adrpo @sjoelund @adeas31 Can you have a look please? This fixes ticket:4407

sjoelund:
Are you sure that works as expected on OSX? From the code, it seems to have a weird behaviour in two cases:
* When the file is writable and can be overwritten, but the file cannot be unlinked (directory read-only, file writable)
* When the file is not writable, but the directory is writable, a file not readable/writable is removed by OMC in order to create the new file

hkiel:
@sjoelund If file/dir permissions are odd then you anyway will run into trouble, so I don't see a (new) problem here.
If unlinking fails, I guess also writing to the old file will fail!? Also, if the file did not exist before, the behaviour should be the same.

hkiel:
@sjoelund I tried the patch on OSX (without any weird file permissions) and it works fine here.

hkiel:
Can we add this patch?

hkiel:
@lochel Do you work with Windows or OSX and experience the same problem?
This patch does not affect Linux.

lochel:
No, sorry. I'm also using Linux...

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-06-16_12-41-07](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4329/console).

PR #1649, opened 2017-05-10
Fix noEventEmit

lochel:
The issue with 'noEventEmit' is that dumping is also prevented for ordinary output points if an event occurs at the same time point. Thus, the result file is empty in the worst case.

### DIFF
--- a/SimulationRuntime/c/simulation/solver/perform_simulation.c
+++ b/SimulationRuntime/c/simulation/solver/perform_simulation.c
@@ -183,7 +183,7 @@ static void fmtInit(DATA* data, MEASURE_TIME* mt)
   }
 }
 
-static void fmtEmitStep(DATA* data, threadData_t *threadData, MEASURE_TIME* mt, int didEventStep)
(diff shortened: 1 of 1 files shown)


lochel:
Can someone please verify this pull request?

SimulationRuntime/c/simulation/solver/perform_simulation.c
```diff
@@ -455,15 +452,15 @@ int prefixedName_performSimulation(DATA* data, threadData_t *threadData, SOLVER_
         syncStep = simulationUpdate(data, threadData, solverInfo);
         retry = 0; /* reset retry */
 
-        fmtEmitStep(data, threadData, &fmt, solverInfo->didEventStep);
+        fmtEmitStep(data, threadData, &fmt, fabs(solverInfo->lastdesiredStep - solverInfo->currentTime) < 1e-12);
```
lochel: If possible, I would like to get rid of this numeric comparison.

PR #1666, opened 2017-05-29
Use nominal value as start value for iteration vars

ptaeuber:
If there is no start value but a nominal value, print a
warning and use the nominal value rather than 0.0

### DIFF
--- a/Compiler/BackEnd/Initialization.mo
+++ b/Compiler/BackEnd/Initialization.mo
@@ -248,7 +248,7 @@ algorithm
     end if;
 
     // warn about iteration variables with default zero start attribute
-    b := warnAboutIterationVariablesWithDefaultZeroStartAttribute(initdae);
+    (initdae, b) := warnAboutIterationVariablesWithDefaultZeroStartAttribute(initdae);
     if b and (not Flags.isSet(Flags.INITIALIZATION)) then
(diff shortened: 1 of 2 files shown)


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-05-29_18-36-26](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4291/).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-05-29_18-57-29](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4292/console).

casella:
I am not sure this is a good idea. From the Modelica specification, section 4.8.1:

_The nominal attribute is meant to be used for scaling purposes and to define tolerances in relative terms_

I don't think we should mix this concept with the start attribute, which has a completely different semantics

ptaeuber:
@casella Yes, this was just a try to fix some models, because I experienced some improvements when I did it manually in the models. This is also already done by some backup heuristics in the solvers as far as I know. I already reverted the commit because the positive effects are because of some other reason:
OpenModelica/OMCompiler#1667
There should be a way to test the coverage without pushing changes in and see what happens over night...

PR #1668, opened 2017-05-30
Do not prefer vars with start value for tearing by default

ptaeuber:
Although it is reasonable to prefer tearing variables with start values, this has the effect that slightly changes in the model (adding start values) lead to different tearing sets, which may behave very unlike. Since this makes analysis and reproducibility hard, I think this flag should not be enabled by default. (Consider the worst case: someone adding a new start value in a model and the model stops simulating.)

Furthermore the positive effect of a prefer by default on the coverage test is not confirmed yet because the coverage summary did not exist when I introduced that flag.

@casella What is your opinion?

### DIFF
--- a/Compiler/Util/Flags.mo
+++ b/Compiler/Util/Flags.mo
@@ -1321,7 +1321,7 @@ constant ConfigFlag DYNAMIC_TEARING_FOR_INITIALIZATION = CONFIG_FLAG(104, "dynam
   NONE(), EXTERNAL(), BOOL_FLAG(false), NONE(),
   Util.gettext("Enable Dynamic Tearing also for the initialization system."));
 constant ConfigFlag PREFER_TVARS_WITH_START_VALUE = CONFIG_FLAG(105, "preferTVarsWithStartValue",
-  NONE(), EXTERNAL(), BOOL_FLAG(true), NONE(),
+  NONE(), EXTERNAL(), BOOL_FLAG(false), NONE(),
   Util.gettext("Prefer tearing variables with start value for initialization."));
 constant ConfigFlag EQUATIONS_PER_FILE = CONFIG_FLAG(106, "equationsPerFile",
   NONE(), EXTERNAL(), INT_FLAG(2000), NONE(),


casella:
In principle, preferring variables with start values is not necessarily a good idea. They may get their start value from some Modelica.SIUnits unit default, which means the modeller hasn't put any special emphasis on them, while they could actually be a bad choice from a numerical point of view.

As I see it, the most preferrable tearing varibles are those from which the function f(x_t) which is solved for x_t in f(x_t) = 0 is "not too much nonlinear". The closer f(x_t) is to linear, the less sensitive the solution process is to the start values (the limit case of linear f(x_t) being totally insensitive to it). Of course defining a measure of "nonlinearity" which can be easily estimated when doing the structural analysis is not at all easy, but this should be the main criterion of choice IMHO.

lochel:
I see only one `Modelica.SIunits` type with default start value and I'm just wondering whether it is a good idea to have start values in there:

```modelica
  type ThermodynamicTemperature = Real (
      final quantity="ThermodynamicTemperature",
      final unit="K",
      min = 0.0,
      start = 288.15,
      nominal = 300,
      displayUnit="degC")
    "Absolute temperature (use type TemperatureDifference for relative temperatures)"
```

casella:
A start value of 0 K is always a very bad idea for an absolute temperature. The rationale behind  setting this  start value was the room temperature was a much better default in many cases. Of course there are some cases where problem-specific start values must be set, but  this default already allows iterative solvers to converge in a number of cases where 0 K would really be problematic, without the need of setting a start attribute, which is in general an annoying thing to do.

hkiel:
One could argue about the value of the temperature. 288.15K=15°C seems to be a special case for aviation (ISA) whereas in physics or chemistry you take either 273.15K=0°C (normal conditions) or 293.15K=20°C (room temperature).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-06-05_17-00-28](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4300/console).

casella:
Sure. It depends a lot on the specific field. For instance, 15 °C is the reference temperature for gas turbine performance maps according to ISO standards.

Anyway, whatever the specific value is, it won't make much of a difference, as long as it is close to room temperature (whatever that means) and far from absolute zero.

PR #1670, opened 2017-06-08
Make Evaluate=false have an effect

ptaeuber:
Parameters with annotation(Evaluate=false) should not be replaced in the system, even if they can be evaluated to constants.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-06-08_11-53-36](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4301/console).

PR #1671, opened 2017-06-12
Match more nonlinear equations first while tearing

ptaeuber:
Until now possible equations for the tarjan matching while tearing are chosen sequentially. These changes choose the most nonlinear equations first to receive preferably linear residual equations.

See also comments on [PR1668](https://github.com/OpenModelica/OMCompiler/pull/1668).

### DIFF
--- a/Compiler/BackEnd/Tearing.mo
+++ b/Compiler/BackEnd/Tearing.mo
@@ -1695,7 +1695,7 @@ author: ptaeuber FHB 2013-2016"
   output Boolean outRunMatching;
 protected
   Integer size, tornsize;
-  array<Integer> ass1, ass2, mapIncRowEqn;
+  array<Integer> ass1, ass2, mapIncRowEqn, eqnNonlinPoints;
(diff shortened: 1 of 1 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-06-12_14-32-58](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4306/console).

casella:
Having residual equations which are linear in the  torn variables is not necessarily good, if the dependencies of those torn variables from the tearing variables (specified in the torn equations) are strongly nonlinear.

ptaeuber:
Sure, you will never get a linear system from a nonlinear one. Even if the residual equations are linear in the torn variables and the torn variables occur linear in all the other equations, the nonlinearity goes down in the residual equations.

Nevertheless, I think it is better to have some heuristics than doing the matching randomly. At least it makes tearing a bit more deterministic regarding model changes.

casella:
I agree with that

PR #1683, opened 2017-06-15
Fix compilation of zmq on OSX

sjoelund:
libzmq uses the LDFLAGS to find libraries, so do not send cmake LDFLAGS
that ignore linker errors since that will give compiler errors on OSX
(the makefile does not check if the symbol exists in a header; it just
compiles a program and checks linking).


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-06-15_09-42-37](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4326/console).

PR #1698, opened 2017-06-25
Remove loop check from differentiation (trac4454)

hkiel:
ther is still a maxIteration check.
reduces running time tremendously

### DIFF
--- a/Compiler/BackEnd/Differentiate.mo
+++ b/Compiler/BackEnd/Differentiate.mo
@@ -140,7 +140,7 @@ algorithm
     funcs := BackendDAEUtil.getFunctions(inShared);
     knvars := BackendDAEUtil.getGlobalKnownVarsFromShared(inShared);
     diffData := BackendDAE.DIFFINPUTDATA(NONE(), SOME(inVariables), SOME(knvars), SOME(inVariables), {}, {}, NONE());
-    (dexp, funcs) := differentiateExp(inExp, DAE.crefTime, diffData, BackendDAE.DIFFERENTIATION_TIME(), funcs, defaultMaxIter, {});
+    (dexp, funcs) := differentiateExp(inExp, DAE.crefTime, diffData, BackendDAE.DIFFERENTIATION_TIME(), funcs, defaultMaxIter);
     (outExp, _) := ExpressionSimplify.simplify(dexp);
     outShared := BackendDAEUtil.setSharedFunctionTree(inShared, funcs);
     if Flags.isSet(Flags.DEBUG_DIFFERENTIATION) then
@@ -183,7 +183,7 @@ algorithm
     if Flags.isSet(Flags.DEBUG_DIFFERENTIATION) then
       BackendDump.debugStrExpStrCrefStr("### differentiateExpSolve\n ", inExp, " w.r.t. ", inCref, "\n");
     end if;
-    (dexp, _) := differentiateExp(inExp, inCref, BackendDAE.emptyInputData, BackendDAE.SIMPLE_DIFFERENTIATION(), fun, defaultMaxIter, {});
+    (dexp, _) := differentiateExp(inExp, inCref, BackendDAE.emptyInputData, BackendDAE.SIMPLE_DIFFERENTIATION(), fun, defaultMaxIter);
     (outExp, _) := ExpressionSimplify.simplify(dexp);
     if Flags.isSet(Flags.DEBUG_DIFFERENTIATION) then
       BackendDump.debugStrExpStr("### Result of differentiateExpSolve\n --> ", outExp, "\n");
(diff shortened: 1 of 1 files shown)


lochel:
I am not sure if we should remove the check completely. However, I agree that the current solution is way too slow according to your analysis. As far as I know, @sjoelund has introduced it to improve the messages that are generated by the max-iteration-check.

I guess it could be implemented more efficiently using hash sets instead of lists, because it is only checked if an element is stored or not.

@sjoelund Any comment?

perost:
You could check the expression stack for duplicates only if the max iteration limit is reached. Then you'd still get the same error message, but the only overhead in case of no error would be building the list of expressions.

lochel:
@perost Sounds like a good idea. On one hand there are cases where this is slower, compared to the hash set approach but on the other hand it should be faster as long as there is no loop.

perost:
@lochel I assumed this was a hard error that would cause the compilation to abort, but otherwise it might be slower letting it run into the iteration limit as you say. But if the compilation is going to abort though it doesn't really matter if that case is slightly slower, having low overhead in the normal case is more important in my opinion.

lochel:
@perost I agree. I just wanted to state it here for the sake of completeness.

hkiel:
A loop in differentiation seems to be a hard error:
```Error.addInternalError("Differentiation failed...");```
So, shall we go for 
a) building the list and checking at iterationMax
or
b) using HashSet?

**Comment by lochel** [comment]
It would be interesting to know how big the overhead is for the hash set approach. If that is too big then I think we should do the check only at iterationMax.

File: Compiler/BackEnd/Differentiate.mo [hunk 1]
```diff
@@ -537,11 +534,6 @@ algorithm
     fail();
   end if;
 */
-  if listMember(inExp, inExpStack) then
```
hkiel: @lochel @perost I see that maximum iteration check is currently disabled

hkiel:
Just for the statistics: not creating the list saves 125MB in ScalableTestSuite.Electrical.BreakerCircuits.ScaledExperiments.BreakerNetwork_N_640 i.e. 15*10^6 elements in the lists...

sjoelund:
Do you think calling GC.free would help?

⁣Sent from BlueMail ​

hkiel:
Not for the list.

hkiel:
Can somebody construct a test case where a loop in differentiateExp() happens???

hkiel:
Can we make a test run on hudson for this PR? Would be interesting to see if it breaks something and if the performance increase can be seen for small models as well.

OpenModelica-Hudson:
Wrong hash OMCompiler e24e32354c5d3613d50833ccc1fbc0b8b6e0b05d e24e323

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-06-28_15-55-07](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4359/console).

hkiel:
I implemented loop check with HashSetExp.
Using HashSetExp instead of list<DAE.Exp> makes no difference both in speed and memory.

So, do we really need the loop checking or can we skip this completely?
@sjoelund commented the maxIteration check out with the following comment:
> This check does not seem to be necessary since looking through the stack of expression seems to stop iteration in most cases, and you get a spam of messages from this check.

Another option would be to add a flag for loop checking (checkDifferentiationLoop). But what would be the default?

lochel:
It would be good to see some examples before drawing a conclusion. @sjoelund, can you point to some examples that trigger this check?

sjoelund:
A HashSet is a very expensive datatype creating a large array by default. You should only use it when you set the size of the set to an appropriate size beforehand. Otherwise, use an AvlSet.

But yes, we could simply have a maximum number of iterations instead as long as it is implemented properly (high enough limit that you don't get messages, and only 1 message per failed differentiation).

sjoelund:
@lochel - this mostly triggered errors when the module was implemented improperly; it's more of a sanity check to see if the compiler is doing something wrong.

hkiel:
@sjoelund if it's only checking for errors in the compiler, then I would skip the loop-check completely, like it is done in this PR.

The maximum iteration number depends on the model size. Currently it is defined as a fixed number (20), but should be set to some pre-calculated number. For ScalableTestSuite.Electrical.BreakerCircuits.ScaledExperiments.BreakerNetwork_N_640 I set it to 700 (640 was too small). I would go for 2048, but then as the function is recursive, I do not know how many iterations the stack can hold...

hkiel:
As @sjoelund stated this check is a sanity check to see if compiler is doing something wrong. So I'm going to merge this PR now.
How can I trigger a run of ScalableTestSuite?

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-07-07_10-00-18](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4382/console).

PR #1730, opened 2017-07-13
Speed up tearing

ptaeuber:
see [ticket:4470](https://trac.openmodelica.org/OpenModelica/ticket/4470)

### DIFF
--- a/Compiler/BackEnd/Tearing.mo
+++ b/Compiler/BackEnd/Tearing.mo
@@ -60,6 +60,7 @@ import ExpressionDump;
 import ExpressionSimplify;
 import ExpressionSolve;
 import Flags;
+import GC;
 import Global;
 import List;
 import Matching;
@@ -1068,12 +1069,12 @@ algorithm
         if Flags.isSet(Flags.TEARING_DUMPVERBOSE) then
           print("Points after 'discriminateDiscrete':\n" + stringDelimitList(List.map(arrayList(points),intString),",") + "\n\n");
         end if;
-    // 4th: Prefer variables with annotation attribute 'tearingSelect=prefer'
+        // 4th: Prefer variables with annotation attribute 'tearingSelect=prefer'
         pointsLst = preferAvoidVariables(freeVars, arrayList(points), tSel_prefer, 3.0);
         if Flags.isSet(Flags.TEARING_DUMPVERBOSE) then
           print("Points after preferring variables with attribute 'prefer':\n" + stringDelimitList(List.map(pointsLst,intString),",") + "\n\n");
         end if;
-    // 5th: Avoid variables with annotation attribute 'tearingSelect=avoid'
+        // 5th: Avoid variables with annotation attribute 'tearingSelect=avoid'
         pointsLst = preferAvoidVariables(freeVars, pointsLst, tSel_avoid, 0.334);
         if Flags.isSet(Flags.TEARING_DUMPVERBOSE) then
           print("Points after discrimination against variables with attribute 'avoid':\n" + stringDelimitList(List.map(pointsLst,intString),",") + "\n\n");
@@ -1717,7 +1718,6 @@ protected
   String s,modelName;
   constant Boolean debug = false;
 algorithm
-
   linear := getLinearfromJacType(jacType);
   BackendDAE.EQSYSTEM(stateSets = stateSets) := isyst;
   noDynamicStateSelection := listEmpty(stateSets);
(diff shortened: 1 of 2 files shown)


hkiel:
The code in [Tearing.mo:4227](https://github.com/OpenModelica/OMCompiler/blob/f50884b47e74fd7770130e9addc22329c9c71ac4/Compiler/BackEnd/Tearing.mo#L4227) can be optimized, because currently
```
powerSet := list(getPowerSetElement(i) for i in 1:(2^size-1));
```
creates a (big!) array for the range and iterates over the array instead of using a for loop. It's better to use this code:

```  powerSet := {};
  for i in (Util.intPow(2,size)-1):-1:1 loop
    powerSet := getPowerSetElement(i)::powerSet;
  end for;
```

This, however needs the new function Util.intPow()
```
public function intPow
  input Integer base;
  input Integer exponent;
  output Integer result = 1;
algorithm
  if exponent >= 0 then
    for i in 1:exponent loop
      result := result * base;
    end for;
  else
    fail();
  end if;
end intPow;
```

ptaeuber:
@hkiel Thanks, I will change this, too. TotalTearing will not be used in practice, though. It was implemented for some theoretical analyses.

hkiel:
@ptaeuber Util.intPow is now checked in

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-07-19_16-17-25](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4416/console).

casella:
@ptaeuber: I've re-run the ScalableTestSuite job after you pushed this. The execution times for tearingSystem in BreakerNetworkDelayed_N_XXX are 0.19 s for N=320, 4.37 s for N=640, and 26.3 s for N=1280. It seems that there is still some scaling problem, at least with this specific model.

ptaeuber:
@casella Yes, I think the scaling is still quite similiar to before, but there should be an overall speed up by factor 7 or something. I will add some more times to compare to [ticket:4470](https://trac.openmodelica.org/OpenModelica/ticket/4470) tomorrow.

ptaeuber:
@casella I get the following times for BreakerNetworkDelayed_N_XXX (O(n^2)):
  320:    0.8249 s
  640:    2.873 s
1280:  10.967 s

Isn't this the latest run of the ScalableTestsuite?:
http://libraries.openmodelica.org/branches/master/ScalableTestSuite/ScalableTestSuite.html

Here the execution times of tearingSystem are similar to the times above.

BreakerNetworkDelayed_N_XXX behaves quite similar to BreakerNetwork_N_XXX, which is the model I compared in detail in [ticket:4470](https://trac.openmodelica.org/OpenModelica/ticket/4470).

casella:
@ptaeuber the link you mention points  to the  run of the ScalableTestSuite that takes  place  automatically with the new testing  infrastructure. The flags used there are not appropriate, one should use the sparse ida solver in ode mode to get  decent simulation performance, and allow for a lot more event iterations, but I cannot change them.

I am still using  the LibraryExperimental  job for my experiments: https://test.openmodelica.org/hudson/job/LibraryExperimental/. Results are reported here https://test.openmodelica.org/libraries/ScalableTestSuite_Experimental/BuildModelRecursive.html. This allows me to set the flags as required.

I have just restarted the job, let's see what we get. It should take about 2 hrs. I guess we should get the same results you mention, regarding the tearingSystem function. Is MC3 now the default method?

hkiel:
@casella Would it make sense to put these simulation settings parameters into some vendor-specific annotation in the model?

casella:
It's not an easy question to answer. For one thing, I want to be able to try different solvers on the library (e.g. ODE mode vs. DAE mode) without  changing  the  annotations everry time. But even more importantly, I would like the tool to be able (eventually) to understand what are the best settings and set itself up. 

Even Simulink does it (https://it.mathworks.com/help/simulink/ug/use-auto-solver-to-select-a-solver.html) and I am sure we can do much better.

ptaeuber:
@casella Yes, MC3 is the default method since 2.5 years :)
https://github.com/OpenModelica/OMCompiler/commit/1112fba0b6e157f452c842518c54e9b148c0185f

casella:
Sorry, I should have RTFM...

casella:
BreakerNetworkDelayed_N
320: 0.33
640: 0.85
1280:  4.62
https://test.openmodelica.org/libraries/ScalableTestSuite_Experimental/BuildModelRecursive.html

PR #1748, opened 2017-07-25
Enable homotopy by default

lochel:
I push this to see the impact on the regression tests.


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-07-25_13-41-42](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4442/).

ptaeuber:
@lochel Almost every failing model fails due to the lambda0 system...

casella:
Unless there are bugs in the elaboration of the lambda0 system, this is an issue with the libraries, not with the tool. 
Dymola by default first tries to directly solve the initial equations and only resorts to homotopy when that fails. The reason is to avoid wasting time with homotopy when not necessary, which is reasonable but also questionable, as homotopy may be essential not only to guarantee convergence, but also to ensure convergence to the right solution. 
Anyway, this may explain why the MSL contains model with singular lambda0 initialization problems.
In the long term, we should fix the MSL models. In the short term, we could add yet another flag (true by default) to try solving the initial system directly without homotopy, e.g. -d=solveWithoutHomotopyFirst

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-07-31_15-44-12](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4450/console).

PR #1762, opened 2017-08-07
Fix bug in function evaluation

ptaeuber:
When predicting the outputs of if branches do not replace
the constants altogether because this leads to wrong results.

This fixes the problem described in https://openmodelica.org/forum/default-topic/2269-problem-with-function.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-08-07_17-09-22](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4464/console).

PR #1767, opened 2017-08-09
Perform incomplete matching for over/underdet syst

sjoelund:
This performs an incomplete matching after producing the message unbalanced
equation system, and instead of just a number differing OMC will try to
perform a sanity check to figure out if for example a variable is never
referenced in any equation or if two variables must be matched by the same
equation.

### DIFF
--- a/Compiler/BackEnd/BackendDAECreate.mo
+++ b/Compiler/BackEnd/BackendDAECreate.mo
@@ -115,6 +115,7 @@ protected
   list<BackendDAE.TimeEvent> timeEvents;
   String neqStr, nvarStr;
   Integer varSize, eqnSize, numCheckpoints;
+  BackendDAE.EqSystem syst;
 algorithm
   numCheckpoints:=ErrorExt.getNumCheckpoints();
   try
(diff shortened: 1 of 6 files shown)


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-08-09_11-12-59](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4469/).

sjoelund:
The unbalanced if-equation seems to get a weird error-message (should it be caught earlier?). Most of the others seem fine.

@lochel @casella should the linked equations in error-messages also contain location information? (OMEdit can't link them, and I don't want to create multiple `From here:` notifications just for the OMEdit link... In messages like these:
```
Error: Too many equations, over-determined system. The model has 10 equation(s) and 8 variable(s).
[simulation/modelica/connectors/HeatTank.mo:32:1-35:15:writable] Notification: Equation 9 (size: 1) inlet.volumeFlowRate = 0.0 cannot solve for enough variables.
  Remaining unsolved variables are: 
  Already solved: inlet.volumeFlowRate
  Equations used to solve those variables:
    Equation 4: inlet.volumeFlowRate = 1.0
[simulation/modelica/connectors/HeatTank.mo:32:1-35:15:writable] Notification: Equation 10 (size: 1) outlet.volumeFlowRate = 0.0 cannot solve for enough variables.
  Remaining unsolved variables are: 
  Already solved: outlet.volumeFlowRate
  Equations used to solve those variables:
    Equation 8: outlet.volumeFlowRate = 2.0
```

You don't see where `inlet.volumeFlowRate = 1.0` is coming from and you can't use `-d=optdaedump` because no equation has been dumped yet. Or should there simply be a debug-flag to dump the equation system during the BackendDAECreate step? (The dumping doesn't show equation sources either though, right?)

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-08-10_12-36-42](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4479/console).

casella:
Ideally, it should be possible to use the declarative debugger.  Is there  enough (or possibly partial) information to get it to work?

I also find the error message a bit odd: "Equation 9 (size: 1) inlet.volumeFlowRate = 0.0 cannot solve for enough variables.". Equations can't _do_ anything, including solving themselves. Maybe "Equation 9 (size: 1) inlet.volumeFlowRate = 0.0 cannot be solved for enough variables." would be more correct?

Last, but not least, it would also be essential to implement Modelica 3.x balanced model checking, see [#3977](https://trac.openmodelica.org/OpenModelica/ticket/3977), otherwise the information provided by partial matching will often be useless (e.g. there is a missing equation in  the following set of 346 equations)

sjoelund:
You can only use the debugger if it can output the debug information, which requires the SimCode structure and thus passing the entire backend.

And yes, the 3.x balanced model checking would be useful (but it's a frontend thing).

Partial matching did find the correct variable never solved for in a set of 4000 equations, but it mostly only works fine if algebraic loops are not involved. It does seem to detect cases where you add an extra equation for something connected though.

PR #1775, opened 2017-08-21
Remove name from class extends SCode.ClassDef.

perost:
- Removed the baseClassName field from SCode.ClassDef.CLASS_EXTENDS,
  since it can never be different from the name already in the class
  element it belongs to.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-08-21_17-57-03](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4492/console).

PR #1777, opened 2017-08-23
evaluateForStatement: do not loop over ranges with stop < start

hkiel:
This concerns bug [trac:3916](https://trac.openmodelica.org/OpenModelica/ticket/3916)

### DIFF
--- a/Compiler/BackEnd/EvaluateFunctions.mo
+++ b/Compiler/BackEnd/EvaluateFunctions.mo
@@ -2253,7 +2253,7 @@ algorithm
 end evaluateFunctions_updateStatement;
 
 
-protected function evaluateForStatement"evaluates a for statement. neste for loops wont work"
+protected function evaluateForStatement"evaluates a for statement. nested for loops wont work"
   input DAE.Statement stmtIn;
   input DAE.FunctionTree funcTreeIn;
   input BackendVarTransform.VariableReplacements replIn;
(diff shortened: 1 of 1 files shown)


sjoelund:
In Modelica, 1:-1 doesn't result in negative steps, so I guess the PR is wrong?

perost:
Looks correct to me, it's returning a step size of -1 if stop < start instead of always returning 1. Note that it returns (start, stop, step), so the order is not the normal (start, step, stop) that you might be used to.

sjoelund:
Yes, it's returning a step size of -1 if stop < start, but it should always return 1. The range `1:-1` is `{}` in Modelica, not `1:0:-1` (like `1:-1:-1`). I'm not really sure why we store the step as an Option instead of adding `DAE.ICONST(1)` if it was not set explicitly...

hkiel:
@sjoelund @vwaurich Can you review this PR please?
The test `simulation/modelica/arrays/Bug3916.mos` has ranges with start=1 and stop=0, so I am not sure wether the range 0:1 (as in the former code) should be looped or the range should be skipped completely (like it is done now, because step size is negative). The result of the test is the same.

hkiel:
Maybe getRangeBounds should always return step=+1 in the first case and the outer function should check that stop>=start?

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-08-23_15-38-09](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4499/console).

PR #1785, opened 2017-08-28
New FrontEnd improvments

mahge:
- Basic implementation of generic/polymorphic functions (function arguments).
  - Implementation of array functions
      - mapping (iterator) functions
      - special Modelica reduction functions.
  - Handling of record Constructors.
  - Basic handling of overloaded operator functions.
      - Expression level handling not implemented yet.
  - Changed CachedData in InstNodes to an array.
      - Each element represents a different type of cache.
      - So far we have caches for functions, package and Inner Outer information.

### DIFF
--- a/Compiler/NFFrontEnd/NFFunction.mo
+++ b/Compiler/NFFrontEnd/NFFunction.mo
@@ -57,6 +57,9 @@ import NFInstNode.CachedData;
 import Lookup = NFLookup;
 import ClassTree = NFClassTree.ClassTree;
 import Prefixes = NFPrefixes;
+import NFLookupState.LookupState;
+import Record = NFRecord;
+
 import MatchKind = NFTypeCheck.MatchKind;
 
 public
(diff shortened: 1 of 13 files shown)


Compiler/NFFrontEnd/NFFunction.mo
```diff
@@ -202,7 +229,9 @@ uniontype Function
     end try;
 
     // Look up the function and create a cref for it.
-    (node, nodes, found_scope) := Lookup.lookupFunctionName(functionName, scope, info);
+    //(node, nodes, found_scope) := Lookup.lookupFunctionName(functionName, scope, info);
```
perost: This will break some of the test cases since assertCallable doesn't handle all the error cases. It would be better to just fix LookupState.assert to handle record constructors correctly instead, and keep all the error checking in one place.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-10-30_11-23-09](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4835/console).

PR #1788, opened 2017-08-30
Fix bug in homotopy solver

ptaeuber:
Do not discard the solution at the beginning of each iteration step.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-08-30_10-53-07](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4512/console).

casella:
Seems like a good idea :)

ptaeuber:
It is not that grave as it may sound :) What is called homotopy solver in the OMC consists of several solvers called in order (newton, hybrid, actual homotopy). So in the most cases the actual homotopy solver (which was affected by the bug) is never used.

casella:
Sure.
I have the feeling, remembering past debugging sessions, that also in other cases the iterative solution process was starting from the original start values instead of from the previous solution. It could be good to check that this doesn't happen anywhere.

PR #1825, opened 2017-09-26
workaround for #4524

hkiel:
the fix should be done in front end -> return type of `size(x,1)` should be integer instead of type of `x`

### DIFF
--- a/Compiler/Template/CodegenCFunctions.tpl
+++ b/Compiler/Template/CodegenCFunctions.tpl
@@ -3667,6 +3667,7 @@ template expTypeFromExpFlag(Exp exp, Integer flag)
  "Generate type helper."
 ::=
   match exp
+  case SIZE(sz=SOME(__))
   case ICONST(__)        then match flag case 8 then "int" case 1 then "integer" else "modelica_integer"
   case RCONST(__)        then match flag case 1 then "real" else "modelica_real"
   case SCONST(__)        then match flag case 1 then "string" else "modelica_string"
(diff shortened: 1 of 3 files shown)


sjoelund:
This is just `expTypeFromExpModelica` returning a bad type, right?

sjoelund:
Just fix line 3694: `case SIZE(__)          then expTypeFlag(typeof(exp), flag)` Should probably be something like:
```mo
  case SIZE(sz=SOME())
  case ICONST(__)        then match flag case 8 then "int" case 1 then "integer" else "modelica_integer"

...
  case SIZE(__) then expTypeFlag(typeof(exp), flag)

  case CONS(__)
  case LIST(__)
  case META_TUPLE(__)
  case META_OPTION(__)
  case MATCHEXPRESSION(__)
  case METARECORDCALL(__)
  case RSUB(__)
  case BOX(__)           then match flag case 1 then "metatype" else "modelica_metatype"
```

sjoelund:
typeof already returns a normal integer for `SIZE(sz=SOME(_))`, so that should be correct...

sjoelund:
Never mind, I see the problem:
```mo
  record SIZE "The size operator"
    Exp exp;
    Option<Exp> sz;
  end SIZE;
```

So yes, `expTypeFromExpFlag` should only be called as it is right now if `sz=NONE()` and the code above I sketched should work (just return integer for sz=SOME()).

hkiel:
thanks, @sjoelund I will try your change.

hkiel:
Other Codegen*.tpl are also affected. I will fix for Cpp and XML.
ADevs, Java and SparseFMI  have no `case SIZE(__)` Why? Shall I add `case SIZE(__)` and `case SIZE(sz=SOME(__))` for them?

sjoelund:
They are pretty much not maintained.

hkiel:
@sjoelund this was a problem with template language (fixed in #1828), so this PR is no longer needed

perost:
I hadn't seen this discussion, so I'd missed that Cpp and XML had the same issue. I'm pushing in the same fix for those too.

PR #1827, opened 2017-09-26
Don't replace dots in fileNamePrefix for FMUs

lochel:
see [ticket:4528](https://trac.openmodelica.org/OpenModelica/ticket/4528)


lochel:
For some reasons the compilation process doesn't work if the target name contains dots:
```
cp: cannot stat ‘me.A.c.so’: No such file or directory
make: *** [me.A.c_FMU] Error 1
```

sjoelund:
The filename has to be identical to modelIdentifier in FMI 1.0, and modelIdentifier must be an allowed C identifier...

lochel:
That means we have to do this hack only for FMI 1.0?

sjoelund:
I believe so.

lochel:
Unfortunately, the FMU generation fails also for FMI 2.0 using this PR.

sjoelund:
Leave all other parts with underscore though to make things simple. Just the name of the FMU with normal dots in it for FMI 2.0 would be nice enough...

rfranke:
FMI 2.0 provides the modelIdentifier as attribute in modelDescription.xml. This way the name of the FMU file is arbitrary. Cpp uses dots in the name of the *.fmu file and underscores for the modelIdentifier, including also binaries/linux64/<modelIdentifer>.so

lochel:
@rfranke Unfortunately, the `fileNamePrefix` attribute of `buildModelFMU(...)` is not considered when using the Cpp runtime. I have reported this in [ticket:4528](https://trac.openmodelica.org/OpenModelica/ticket/4528). Can you have a quick look?

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-09-28_12-22-51](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4591/).

lochel:
@sjoelund The fmu export works now as expected, but the fmu import doesn't. It is expected that the generated Modelica-wrapper's name is `<%fmiInfo.fmiModelIdentifier%>_<%getFMIType(fmiInfo)%>_FMU`. So there are no dots allowed. How should we handle this?

sjoelund:
I never worked on the FMI import. Maybe @adeas31 has some idea on how to do it. But the name of the Modelica model should of course be allowed to not contain dots. It is the filename used by the importer that matters (and it needs to know the filename might be different from the identifier).

sjoelund:
@rfranke @lochel I agree the C++ FMUs need to respect the fileNamePrefix. It's a hard pre-requisite for the coverage testing (with the current implementation, you get race conditions when running the coverage suite if you enable C++ FMUs)..

sjoelund:
1e02c69fc976e6586d1e1d45b4368715019c613d made the C++ runtime accept fileNamePrefix as expected by the API functions.

sjoelund:
@lochel I guess for the generated variable name, use single quotes or just replace dots with underscore. It's just a generated internal name after all, right? (Which needs to be updated in other locations in the importer)

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-10-02_13-00-17](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4607/).

lochel:
The FMU's filename may now contain dots for FMI 2.0. However, the wrapper model name doesn’t.
For example:
```
translateModelFMU(Modelica.Mechanics.MultiBody.Examples.Elementary.Pendulum, version="2.0");
importFMU("Modelica.Mechanics.MultiBody.Examples.Elementary.Pendulum.fmu");
loadFile("Modelica_Mechanics_MultiBody_Examples_Elementary_Pendulum_me_FMU.mo");
```
Note the underscores in the last line.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-10-02_13-22-36](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4608/console).

PR #1841, opened 2017-09-28
optimize merge sort for memory usage

hkiel:
saves about 25% of the used memory at approx. same time


hkiel:
makes code more complicated than it is worth it.

sjoelund:
If you don't want to touch already sorted lists, it is of course possible to check if each subtree is already sorted to avoid some allocations without much added complexity

hkiel:
This PR was not about sorted lists, but reducing memory overhead of lists, but that is only a minor impact. I checked with ScalableTestSuite models which use some GB, and could reduce memory usage by approx. 10MB

PR #1846, opened 2017-09-30
Fix for OperatorOverloading.makeEnumOperator.

perost:
- Change makeEnumOperator so it doesn't return both types when only the
  second is an enumeration, so that type checking of operations doesn't
  always succeed as long as the second operand has an enumeration type.


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-09-30_15-53-56](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4600/).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-09-30_16-14-11](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4601/console).

PR #1847, opened 2017-10-01
silence static analyzer warning for dead assignment in match expressions

hkiel:


hkiel:
@sjoelund @adrpo what do you think? Is it worth it?
I would estimate that about 1000 warnings will go away with this.

sjoelund:
I'm thinking about another way to fix this: by using labels and goto, and just avoid having this variable at all. I'll see if I can make that work; should make the code perhaps a bit more compact.

sjoelund:
Fixed by #1857

PR #1850, opened 2017-10-02
Fix generated C++ FMU filenames

sjoelund:
Previously, the API reported dots were used, but underscores were
actually being used. The problem is that the FMI templates changed
the meaning of the fileNamePrefix (from the OM API user's perspective).


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-10-02_14-53-50](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4612/console).

PR #1854, opened 2017-10-03
Make getAnnotationNamedModifiers not return bad data

sjoelund:
If the annotation does not exist, use the error buffer to return the
message instead of saying an annotation with a weird name exists.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-10-03_11-09-56](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4621/console).

PR #1857, opened 2017-10-04
[MetaModelica] Use goto instead of done variable

sjoelund:
By using goto directly in match expressions, we avoid the static
analyzer telling we have an unused assignment to variable `done` in some
cases.

This fixes #1847 in an alternative way.

### DIFF
--- a/Compiler/Template/CodegenCFunctions.tpl
+++ b/Compiler/Template/CodegenCFunctions.tpl
@@ -6740,11 +6740,10 @@ case exp as MATCHEXPRESSION(__) then
           <%varDeclsInner%>
           <%preExpInner%>
           <%match exp.matchType
-          case MATCH(switch=SOME(_)) then '<%done%> = 0;<%\n%>{'
+          case MATCH(switch=SOME(_)) then '{'
           else
           <<
           <%ix%> = 0;
-          <%done%> = 0;
           <% match exp.matchType case MATCHCONTINUE(__) then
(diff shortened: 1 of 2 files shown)


OpenModelica-Hudson:
The job failed to compile; for details, see [OpenModelica_TEST_PULL_REQUEST 2017-10-04_13-02-07](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4626/console).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-10-04_14-09-26](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4627/console).

hkiel:
@sjoelund The impact of this change is big, but the wrong way. About 1000 warnings went away just to be replaced by 3500 new, e.g.:
```
tmp4 += 11; /* Pattern matching succeeded; we may skip some cases if we fail */

 -> Value stored to 'tmp4' is never read
```

sjoelund:
There were 0 of those in MetaUtil.mo that I tested on :(

sjoelund:
So... those are for matchcontinue cases. And that statement that the value is never read is actually... wrong. Because there might be a longjmp in a function called in those statements...

hkiel:
Sure, but clang static analyzer finds this as a warning, maybe because it does not know about longjmp?

sjoelund:
#1863 should fix this.

PR #1858, opened 2017-10-04
Fix for fileNamePrefix not a valid C identifier

sjoelund:
Previously, we only handled dots in the prefix; now we replace all
invalid characters in C identifiers with an underscore.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-10-04_14-48-09](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4629/console).

PR #1859, opened 2017-10-04
inner/outer handling for nfinst.

perost:
- Implemented handling of inner/outer, including automatic generation
  of inner elements when missing.
- Moved setting the type of a component from instComponent to
  instClass, to allow lookup in a component's parent component
  during instantiation.
- Fixed scoping of builtin variables (i.e. time), to avoid them
  being incorrectly prefixed.
- Fixed lookup of self-referencing extends and added check for
  inherited base classes.

### DIFF
--- a/Compiler/FrontEnd/SCode.mo
+++ b/Compiler/FrontEnd/SCode.mo
@@ -3871,6 +3871,24 @@ algorithm
   end match;
 end isNotBuiltinClass;
 
+public function getElementNamedAnnotation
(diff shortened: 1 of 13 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-10-04_17-47-26](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4631/console).

PR #1863, opened 2017-10-05
Avoid static analyzer dead assignment warning

sjoelund:


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-10-05_17-00-18](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4638/console).

hkiel:
@sjoelund This is about the same solution I supplied with the original problem ;-)
I hope that dumb compilers (msvc?) do not add a NOP for this...

sjoelund:
The difference is that this is only done for matchcontinue which is anyway something we want to avoid. Anyway, having fewer variables is good because it means fewer things for setjmp/longjmp to store. I much prefer doing goto over having weird variables that might be clobbered by a longjmp (though I haven't seen any such bugs in a long time)

PR #1869, opened 2017-10-08
Change the default simulation opt. level to -Os

sjoelund:
Testing has shown that the performance of clang seems to be good enough
to enable -Os by default now. It's still slightly slower for most
models, especially small ones. But it should be a good default for
larger models.

The optimization level for simulations is now set by Susan, and
setCFlags will override this. This makes it possible for dynload to use
a different optimization level unless set by setCFlags. The disadvantage
is that getCFlags does not show what optimization level the simulations
will use, and this will be different for C++ and C.

Note that this is how the MinGW version of OM worked before (except the
default for simulations was -O0).

OMEdit also has an option to change optimization level, so it should be
reasonably convenient for the user to change this to -O0 if desired.


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-10-08_08-08-07](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4647/).

sjoelund:
About a minute slower for the testsuite using -Os for simulation compilation... And some models get slightly different results. @lochel @adrpo, what do you think? Try with -O1 or -O2? I wouldn't want to have special opt-level for rtest and for the user, the most parallel part of the compiler is compiling the source code so from a user's perspective the optimization flags should probably help...

hkiel:
I guess for the OMEdit debugger, we should also always set -O0 ?

sjoelund:
We do. It sets the default to -O0 -g instead of the -Os I set

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-10-10_11-11-53](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4673/).

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-10-10_11-52-28](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4675/).

sjoelund:
-O1 has pretty much the same performance as -Os

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-10-11_18-24-44](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4710/).

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-12-05_14-01-07](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5013/).

sjoelund:
My new changes use -O0 when running the testsuite so we get the same old testing but better benchmarks.

sjoelund:
I also disabled -march=native since it is annoying for cross-compilation and when generating FMUs.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-03-14_06-46-47](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5406/console).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-03-14_07-03-00](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5407/console).

PR #1894, opened 2017-10-13
Fix crash is default linear solver

sjoelund:
If an expression makes a longjmp, the linear solver data was not
restored. This fix makes it so the solver data does not need to be
patched depending on which linear solver is running; it simply allocates
memory for an extra void* to hold the fallback solver's data.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-10-13_07-46-16](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4726/console).

PR #1904, opened 2017-10-14
fix Matching.removeEdgesForNoDerivativeFunctionInputs

hkiel:


Compiler/BackEnd/Matching.mo
```diff
@@ -5666,7 +5666,7 @@ algorithm
       arrayUpdate(m,idx,row);
       //update mt
       for varIdx in varIdxs loop
-        row := arrayGet(m,varIdx);
+        row := arrayGet(mt,varIdx);
```
hkiel: @sjoelund Is this fix correct? From my understanding mt should be updated here, but maybe it is the other way round!?

hkiel: @vwaurich Can you have a look please?

sjoelund: I'm not sure. I guess it's reasonable to get the value and update it instead of getting some different value...

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-10-26_07-49-54](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4816/console).

PR #1919, opened 2017-10-17
Fix hash tables

lochel:
The `HashVector` contains in some cases indices that point outside the range of `ValueArray`. As far as I can see, this is due to `function bucketToValuesSize`.

Does anyone know why that function is actually used?

https://github.com/OpenModelica/OMCompiler/blob/f55d7d6b591207f514adfdca310c7b858f7e141d/Compiler/Util/BaseHashTable.mo#L99-L105

### DIFF
--- a/Compiler/Util/BaseHashTable.mo
+++ b/Compiler/Util/BaseHashTable.mo
@@ -96,14 +96,6 @@ partial function FuncEq input Key key1; input Key key2; output Boolean b; end Fu
 partial function FuncKeyString input Key key; output String str; end FuncKeyString;
 partial function FuncValString input Value val; output String str; end FuncValString;
 
-public function bucketToValuesSize
-"calculate the values array size based on the bucket size"
-  input Integer szBucket;
-  output Integer szArr;
-algorithm
-  szArr := realInt(realMul(intReal(szBucket), 0.6)); // intDiv(szBucket, 10);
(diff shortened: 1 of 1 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-10-17_22-55-01](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4767/console).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-10-18_09-52-26](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4768/console).

sjoelund:
One of them is the number of possible entries in the hash table (e.g. 100 buckets to drop things into; but the values is the maximum number of entries in the hash-table, which might then be 60; so you have a maximum of 60 entries sharing 100 hash values). Use an AvlTree for small sets; it's much easier to use! Or perhaps pass an additional input (default=0 means use the previous default and otherwise you can select how many entries you want at most in the table).

lochel:
I did some more debugging and the use of that function is fine. The problem is that the hash tables are partially mutable, which means that the entire data structure may get corrupted in match continue statements.

PR #1923, opened 2017-10-18
Unit checking crashes because of the hash tables

lochel:
In some cases, the unit checking crashes. This is probably due to side-effects in the hash tables that are used.


sjoelund:
Could HashTableStringToUnit be removed as well?

lochel:
Sure, but I want to take one step after another ;)

lochel:
We have also to do the same changes to the ported NF implementation...

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-10-18_14-51-07](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4778/).

lochel:
For some weird reason, this changes the result of the unit checking in some cases.

PR #1928, opened 2017-10-19
Add MetaModelica.Dangerous.listArrayLiteral

sjoelund:


sjoelund:
@perost - check if you need a new bootstrapping tarball due to this. The Susan template is a bit annoying to patch, but if it's not needed for bootstrapping due to not being called during bootstrapping, I could probably patch something in.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-10-19_17-24-18](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4794/console).

perost:
@sjoelund A new bootstrapping tarball seems to be needed, yes. The new instantiation isn't really needed for bootstrapping though, so maybe you can exclude it somehow. Also, I assume b5d86d3277e30ebb4c9222de2379af53e8029b86 won't work without a new tarball either. Not terribly important, but would be nice to get that working too.

sjoelund:
I tested https://github.com/OpenModelica/OMCompiler/pull/1929 with bootstrapping and it seems to work (the literals will actually be `listArray` calls during bootstrapping and thus just slower and mutable). b5d86d3 could probably be fixed with a patch to the bootstrapping sources.

PR #1943, opened 2017-10-25
Use sets in comSubExp

sjoelund:
This improves performance of comSubExp (ticket:4398). Part of the
underlying problem is that the adjacency matrix stores everything as
list<Integer>, but really is a set of integers. Using List.intersect,
and similar functions is much slower than using for example AvlSetInt.

PR #1944, opened 2017-10-25
Use sets instead of lists in comSubExp

sjoelund:
This resolves some performance problems reported in ticket:4398.

### DIFF
--- a/Compiler/BackEnd/ResolveLoops.mo
+++ b/Compiler/BackEnd/ResolveLoops.mo
@@ -40,8 +40,8 @@ import BackendDAE;
 import DAE;
 
 protected
-
 import Array;
+import AvlSetInt;
 import BackendDAEUtil;
 import BackendEquation;
 import BackendVariable;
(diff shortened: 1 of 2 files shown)


sjoelund:
This is the second half of #1943

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-10-25_17-08-16](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4815/console).

hkiel:
I remember I added code to sort the lists to speed up comparisons. Is that patch still needed???
https://github.com/OpenModelica/OMCompiler/commit/9a5c35eeed772b0c4ab63acfdcdec65b76f2da81

sjoelund:
I only removed parts of it (because a sorted list is not as fast as a binary tree for lookup / intersection; mainly the intersection on the sorted list created a temporary list that was not needed)

hkiel:
So the remaining parts of the sorting must stay?

sjoelund:
I think so. See also https://trac.openmodelica.org/OpenModelica/ticket/4589 for my proposal on how to help write better backend modules

PR #1950, opened 2017-10-27
Disable CC optimizations on linear systems

sjoelund:
Linear systems are huge and typically just assign array elements to
literal values. Huge functions take a long time to optimize for little
gain (relative to the cost of running the linear solver).


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-10-27_13-30-22](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4823/console).

casella:
Just as a side comment, linear systems are not huge most of the time, they can be in some special classes of systems, like the ones that we put on purpose in the ScalableTestSuite.

That said, if they are small there are even less reasons to optimize them, so I support this PR.

PR #1955, opened 2017-10-30
Added optional Susan debugging help

sjoelund:
There is a new flag -d=susanDebug which can generate a different
MetaModelica from the Susan code. It will give a stack-trace of the
point where a failure happens.

Use the debug-flag when compiling the Susan code to add extra try/catch
blocks everywhere (this slows down the code considerably and is not
intended for use in the production executable; only for debugging).

### DIFF
--- a/Compiler/Template/Tpl.mo
+++ b/Compiler/Template/Tpl.mo
@@ -8,16 +8,18 @@ encapsulated package Tpl
   $Id$
 "
 
-protected import Config;
-protected import ClockIndexes;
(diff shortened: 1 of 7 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-10-30_09-37-33](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4834/console).

PR #1961, opened 2017-11-01
Fix scalability for systems without variables

sjoelund:
If a system has only parameters and no equations, we now properly
estimate the size of the HashSet used for creating the ModelInfo
structure as variables+parameters. This avoids using linear search
due to the HashSet having size 1. This fixes ticket:4590.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-11-01_07-46-45](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4845/console).

PR #1962, opened 2017-11-01
Use ZeroCrossingSet instead of searching in list

sjoelund:
This fixed the main issue reported in ticket:4584.

### DIFF
--- a/Compiler/BackEnd/BackendDAE.mo
+++ b/Compiler/BackEnd/BackendDAE.mo
@@ -594,7 +594,7 @@ uniontype EventInfo
   record EVENT_INFO
     list<TimeEvent> timeEvents         "stores all information related to time events";
     ZeroCrossingSet zeroCrossings "list of zero crossing conditions";
-    DoubleEndedList<ZeroCrossing> relations    "list of zero crossing function as before";
+    ZeroCrossingSet relations    "list of zero crossing function as before";
     ZeroCrossingSet samples       "[deprecated] list of sample as before, only used by cpp runtime (TODO: REMOVE ME)";
     Integer numberMathEvents           "stores the number of math function that trigger events e.g. floor, ceil, integer, ...";
   end EVENT_INFO;
--- a/Compiler/BackEnd/BackendDAECreate.mo
+++ b/Compiler/BackEnd/BackendDAECreate.mo
@@ -147,7 +147,7 @@ algorithm
   eqnarr := BackendEquation.listEquation(eqns);
   reqnarr := BackendEquation.listEquation(reqns);
   ieqnarr := BackendEquation.listEquation(ieqns);
-  einfo := BackendDAE.EVENT_INFO(timeEvents, ZeroCrossings.new(), DoubleEndedList.fromList({}), ZeroCrossings.new(), 0);
+  einfo := BackendDAE.EVENT_INFO(timeEvents, ZeroCrossings.new(), ZeroCrossings.new(), ZeroCrossings.new(), 0);
   symjacs := {(NONE(), ({}, {}, ({}, {}), -1), {}), (NONE(), ({}, {}, ({}, {}), -1), {}), (NONE(), ({}, {}, ({}, {}), -1), {}), (NONE(), ({}, {}, ({}, {}), -1), {})};
   syst := BackendDAEUtil.createEqSystem(vars_1, eqnarr, {}, BackendDAE.UNKNOWN_PARTITION(), reqnarr);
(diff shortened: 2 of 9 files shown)


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-11-01_09-00-04](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4846/).

hkiel:
@sjoelund Do you have any idea on how to fix this?

sjoelund:
No, I'm not really sure how to fix it, or if it's even a problem.

hkiel:
When unconditionally adding all ZeroCrossings in `ZeroCrossings.add` by commenting out the `if not contains(zc_set, zc) then` and the following `arrayUpdate()`, all tests in `testsuite/simulation/libraries/3rdParty/PNlib` work again.

Either these tests rely on duplicate occurrence of some ZeroCrossings or there are some ZeroCrossings that seem to be the same, but are not.

When I just comment out the `if not contains(zc_set, zc)` then I get
`Error: post-optimization module findZeroCrossings (simulation) failed.`

hkiel:
The problem comes from the `add_list` function.
In the old version for `Test3.mos` we have
`data->modelData->nRelations = 4;`
and with your patch we get
`data->modelData->nRelations = 3;`

hkiel:
add is not adding new ZC because it already exists in the set (old):
`BackendDump.zeroCrossingString` says:
old: `T1.activation.tOut[1] >= T1.activation.maxTokens[1] with index = 0 in equations [4]`
new: `T1.activation.tOut[1] >= T1.activation.maxTokens[1] with index = 1 in equations [4]`
ExpressionDump.printExpStr says:
old:  `T1.activation.tOut[1] >= T1.activation.maxTokens[1]`
new: `T1.activation.tOut[1] >= T1.activation.maxTokens[1]`

@sjoelund  Do you know what `with index = 0` means? Does this zc have a wrong number, so that it is not generated later on? Are they different?

Maybe we should use `BackendDump.zeroCrossingString` as key instead of `ExpressionDump.printExpStr`?

wibraun:
@hkiel `with index = 0` means the ZeroCrossing contains a relation with index 0. And yes, they have to start with 0.

I haven't look in to the code, but it looks like the issue from ticket:3358 raised again, that zeroCrossings in algorithm are merge, what is wrong.

wibraun:
Actually, we should redesign the zero-crossing handling at all, according to #2061.

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-11-30_14-28-24](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5007/).

OpenModelica-Hudson:
The job failed to compile; for details, see [OpenModelica_TEST_PULL_REQUEST 2019-04-11_09-00-51](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6673/console).

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2019-04-11_11-49-23](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6675/).

hkiel:
@sjoelund Do you have an idea if the regressions are really problems?

sjoelund:
Well, they sort of are a problem:
```
 |                 | |       | | [1] (pre:  true)  true = h <= 0.0
 |                 | |       | | [2] (pre: false) false = v <= 0.0
-LOG_EVENTS_V      | info    | status of zero crossings at time=2.55860059936
+| | |       | | [3] (pre:  true)  true = h <= 0.0
+| | |       | | [4] (pre: false) false = v <= 0.0
```

Says before there were 2 event indicators and now there are 4. So simulations become slower. Also, these event indicators are identical to other ones...

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2019-04-11_13-47-22](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6676/).

hkiel:
The problem is that with the existing code the same relation can be added multiple times, because they are simply added to the list. When switching to ZeroCrossingSet these duplicate relations are no longer added. This can be "fixed" by considering relations with different index as different (as in my path) despite having the same expression, but that causes problems in the existing code.

So the current code is already "broken", as it depends on having the the same relation multiple times in the list.

PR #1968, opened 2017-11-01
Add flag to fix the mat file headers every N steps

lochel:
This enables reading of partial simulation results while the simulation is actually not finished.
It is disables by default, because it slows down the simulation (especially on slow hard drives).

Usage: `model.exe -mat_sync=<N>`

**TODO**
- [x] Test


lochel:
Tested with OMPlot on Linux and it works nicely. I cannot test it on Windows, but I guess it works there as well.

lochel:
It works also with Dymola and [MC's PlotXY](http://www.dsea.unipi.it/Members/ceraolow/Software/mcs-plotxy/).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-11-02_23-47-55](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4863/console).

lochel:
@adeas31 Can we have an option in OMEdit/OMPlot to (re-)load the partially written mat file every N seconds?

sjoelund:
That would be quite annoying and resource-consuming. Perhaps you want a functionality where you resume reading the file from the previous position. It requires some substantial changes though.

lochel:
Well, that wouldn't be turned on by default. But in fact, that would be rather nice. I think even a simple reload-button would be good enough.

adeas31:
@lochel adding a reload button to plot window doesn't make sense since the plot window can have variables plotted from several result files. We could add a reload button to a context menu of a variable.

In OMEdit you can uncheck and recheck the variable in order to reload/replot it. In OMPlot just call the executable again and it will update the curve.

lochel:
@adeas31 How can I plot signals from different result files into the same OMPlot windows?

adeas31:
Well calling the executable twice like this,
```
./OMPlot.exe --filename=A_res.mat a &
./OMPlot.exe --filename=B_res.mat b &
```
should show two curves but right now it is clearing the old curve before plotting the new one.

adeas31:
@casella and @lochel we could have a refresh button with a drop down. In the drop down we list all the result files whose variables are plotted in the window. The user selects the file to reload and the variables associated to it are updated. What do you think?

lochel:
The refresh button could just reload all the files of the given plot window.

adeas31:
Can you make an enhancement ticket?

lochel:
Now, there is a ticket for the discussion: [ticket:4609](https://trac.openmodelica.org/OpenModelica/ticket/4609)

casella:
[`My proposal`](https://trac.openmodelica.org/OpenModelica/ticket/4609#comment:1)

casella:
A well known competitor in the simulation world (I'm not talking about Modelica tools) introduced the "scope" concept, about 25 years ago. I think that concept is still valid :)

PR #1982, opened 2017-11-06
Improve element visibility in nfinst.

perost:
- Move visibility to InstNode so that it's always available for any
  class or component.
- Fix attribute propagation so that components don't inherited
  visibility from their parents.
- Turn on visibility check in lookup for classes (fixes ticket:2296).
- Fixed stream connection handling having undefined behaviour
  due to ticket:4502.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-11-06_17-20-57](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4882/console).

PR #1987, opened 2017-11-07
NFFrontEnd record fix

mahge:
- create a typed component for the return type of default constuctors.
  - when converting a function to old DAE structures use the splitted inputs, outputs and locals instead of the components from the class node so that the auto generated return component is picked up. It is not added to the classInst.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-11-07_11-52-24](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4888/console).

PR #2010, opened 2017-11-13
Improve detection of invalid each in NFInst.

perost:
- Take the dimensions of a component's type into account when checking
  whether each is used on an array or not.
- Don't fail when an invalid use of each is found, since the MSL
  uses each incorrectly.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-11-13_17-15-51](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4927/console).

PR #2016, opened 2017-11-14
NFInst improvements.

perost:
- Handle enumeration relations in NFSimplifyExp.
- Fix case for if-equation in NFConvertDAE.convertInitialEquation
  so that it generates an initial if-equation instead of a normal one.
- Fix NFScalarize.scalarizeIfEquation to handle the case where an
  if-equation is completely removed due to all conditions being false.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-11-14_16-07-32](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4935/console).

PR #2017, opened 2017-11-15
Partially type-check for min,max,smooth

sjoelund:
Does not handle all cases yet (min/max does not check if `<` or `>`
can be used on the expressions; smooth does not check for all allowed
expressions, like records containing only arrays of reals, etc).


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-11-15_09-20-31](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4936/console).

PR #2031, opened 2017-11-17
[CodeGenC] skip cases independent of passing guard

hkiel:
if cases can be skipped in a match expression, then do it independent of passing the guard


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-11-17_10-10-14](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4960/console).

lochel:
Did you notice any impact on the performance?

hkiel:
Only marginally: 282s -> 276s for `ScalableTestSuite.Elementary.Tables.ScaledExperiments.CombiTimeTable_N_4000`
I expected more, but that is not true because advancing is only done for a complete match (see [ticket:4633](https://trac.openmodelica.org/OpenModelica/ticket/4633)) and not for partial matches which would be far more relevant.

PR #2032, opened 2017-11-17
fix "a or not a" and "a and not a" simplifications

hkiel:
optimize some simplify rules
reuse existing structures

PR #2034, opened 2017-11-19
workaround for PNLib

hkiel:
in PNlib the `DAE.LBINARY` expression `{} and not {}` must be handled and should not be simplified to `false` by `ExpressionSimplify.simplifyLBinary`

### DIFF
--- a/Compiler/FrontEnd/ExpressionSimplify.mo
+++ b/Compiler/FrontEnd/ExpressionSimplify.mo
@@ -4777,6 +4777,11 @@ algorithm
       Real r;
       Option<DAE.Exp> oexp;
 
+    // fix for PNLib where "{} and not {}" is evaluated
+    case (_,_,DAE.ARRAY(array={}),_)
+      then origExp;
+    case (_,_,_,DAE.ARRAY(array={}))
+      then origExp;
     // a AND not a -> false
     case (_,DAE.AND(_),e1,DAE.LUNARY(DAE.NOT(_),e2))
       guard Expression.expEqual(e1, e2)


hkiel:
workaround for #2032

sjoelund:
Wouldn't it be better to check for array type in the other cases? You could have crefs which are arrays and here wrong code still

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-11-19_08-48-31](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4966/console).

hkiel:
I now removed the extra cases again and changed operator check to `DAE.AND(DAE.T_BOOL())` and  `DAE.OR(DAE.T_BOOL())` for the cases where a `DAE.BCONST()` is created.
That should fix the problem in a good manner!?

Compiler/FrontEnd/ExpressionSimplify.mo
```diff
@@ -4777,6 +4777,11 @@ algorithm
       Real r;
       Option<DAE.Exp> oexp;
 
+    // fix for PNLib where "{} and not {}" is evaluated
+    case (_,_,DAE.ARRAY(array={}),_)
+      then origExp;
+    case (_,_,_,DAE.ARRAY(array={}))
+      then origExp;
```
hkiel: @sjoelund Would it be okay to return the empty arrays in these two cases (instead of `origExp`)?
This would reduce the `{} and not {}` of PNLib to `{}`
I could add these cases again, if that is okay. I don't know if these rules would shadow some error or so.

PR #2037, opened 2017-11-20
[NF] Handle recursive functions.

perost:
- Update the function cache before instantiating a function's
  expressions, to avoid instantiating it again in the case of
  a recursive call.
- Split the typing of functions so their signatures are typed
  first, then type their bodies only after updating the cache.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-11-20_16-09-12](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4969/console).

PR #2048, opened 2017-11-23
 implemented heap sort for arrays

hkiel:
reduces memory usage in `ScalableTestSuite.Electrical.DistributionSystemAC.ScaledExperiments.DistributionSystemLinearIndividual_N_28_M_28` from 950MB to 280MB in `simCode: created linear, non-linear and system jacobian parts`

### DIFF
--- a/Compiler/SimCode/SimCodeUtil.mo
+++ b/Compiler/SimCode/SimCodeUtil.mo
@@ -5002,6 +5002,7 @@ protected
   DAE.ComponentRef cref;
   Integer size, i, j;
   list<Integer> intLst;
+  array<Integer> intArr;
   list<DAE.ComponentRef> crefs;
 algorithm
   //create HT
@@ -5028,7 +5029,9 @@ algorithm
          j := BaseHashTable.get(cr, ht);
          intLst := j :: intLst;
        end for;
-       intLst := List.sort(intLst, intGt);
+       intArr := listArray(intLst);
+       Array.heapSort(intArr);
+       intLst := arrayList(intArr);
        outSparse := (i, intLst) :: outSparse;
     end for;
     outSparse := List.sort(outSparse, Util.compareTupleIntGt);
@@ -5045,6 +5048,7 @@ protected
   DAE.ComponentRef cref;
   Integer size, i, j;
   list<Integer> intLst;
+  array<Integer> intArr;
   list<DAE.ComponentRef> crefs;
 algorithm
   //create HT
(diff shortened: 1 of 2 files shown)


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-11-23_17-54-22](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4981/).

hkiel:
Argh! Some tests rely on sort being the way it is now (= merge sort) and not being a stable sort algorithm.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-11-24_11-12-32](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/4982/console).

hkiel:
@casella @sjoelund Can you check the impact on ScalableTest suite please?
Now, sorting should use O(n) memory instead of O(n*log(n)) which should increase performance for large models.
We can apply this sort algorithm on more integer sorts (see #2050) and optionally on other types as well, given that the result of the sorting accepts the result of a stable sort.

casella:
@hkiel, I misunderstood that you have tested this on the new front-end, so I started a run of the NF tests on Hudson had a look at the results using it. Of course the model you mentioned doesn't reach the SimCode phase because it uses Complex, so the NF fails outright. The only distribution systems that work all the way down to code generation with the NF are ScalableTestSuite.Electrical.DistributionSystemDC.ScaledExperiments.DistributionSystemModelica_N_XX_M_XX.

I checked the larger one N = M = 112, and I saw that indeed the memory allocation has reduced from 3.98 GB to 1.886 GB (ratio 2.11), though the time has remained roughly the same, about 15 s.

For N = M = 28 (which is one sixteenth of the dimension of the previous one) the size went down from 243 to 108 MB (ratio 2.25); the time went from 0.71 to 0.63, so it's still roughly the same. If the complexity had really gone from O(N log(N)) to O(N) I would have expected to see a larger ratio of improvement for the larger model, but maybe we haven't still reached the asymptotic value. 

Anyway, a factor 2+ reduction in memory consumption seems to be very good. N = M = 112 has a quarter million equations, so if we save 2 GB that means 8 kB per equation, which is not bad. 

I would suggest to try applying this elsewhere and see what happen. When we have a new run of Hudson with the old front-end I can also check on other models, but I expect to see more or less the same result, since the improvement is not on the front-end.

PR #2061, opened 2017-11-30
Redesign ZeroCrossing handling ( ticket:3851)

wibraun:
- introduce new Expression for zero crossings
- this will make the search of ZeroCrossings much easier, because currently we write the index of a 
zeroCrossing in the the involved Relation, but need to collect them both.


hkiel:
Any progress on this topic?

wibraun:
Unfortunately no :( there are higher prioritized tasks on my todo, and this has a nice to have status.

PR #2074, opened 2017-12-13
Some work-arounds for Ceval of zero-size arrays

sjoelund:
The Values.mo structure does not store the type of the array which means
that `Values.ARRAY({})` becomes `{} /* T_UNKNOWN[0] */` instead of
containing the actual type. We now have some work-arounds for this case:
- cevalIfConstant will not try to evaluate literal values
- If an empty array is returned in cevalIfConstant, we backpatch the
  array type to the expected type (but this only works for the top-level
  expression; hopefully this is enough to not propagate unknown
  dimensions too far though).

### DIFF
--- a/Compiler/FrontEnd/Ceval.mo
+++ b/Compiler/FrontEnd/Ceval.mo
@@ -885,79 +885,63 @@ public function cevalIfConstant
    or if the expression is a call of parameter constness whose return type
    contains unknown dimensions (in which case we need to determine the size of
    those dimensions)."
-  input FCore.Cache inCache;
+  input output FCore.Cache cache;
   input FCore.Graph inEnv;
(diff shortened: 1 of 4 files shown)


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2017-12-13_08-23-14](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5026/).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-12-13_09-09-56](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5027/console).

PR #2077, opened 2017-12-13
[NF] Fix external object calls.

perost:
- Change external object calls so they call the constructor instead.
- Use an empty class tree for external object so the constructor or
  destructor can't be called explicitly.
- Fix error message for when a function can't be found to say function
  instead of variable.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-12-13_16-09-33](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5036/console).

PR #2080, opened 2017-12-15
[NF] Use Error.assertion() instead of assert()

sjoelund:
Error.assertion reports errors to the error buffer, which means that the
message can be access in scripting and OMPython.

This fixes ticket:4676


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-12-15_07-19-06](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5039/console).

PR #2084, opened 2017-12-16
improve performance of ExpressionSimplify

hkiel:
avoid excessive longjmp (matchcontinue)
avoid recursion (stack overflow) when processing lists

### DIFF
--- a/Compiler/FrontEnd/Expression.mo
+++ b/Compiler/FrontEnd/Expression.mo
@@ -1525,17 +1525,18 @@ public function realExpIntLit "returns the int value if expression is constant R
   input DAE.Exp exp;
   output Option<Integer> oi;
 algorithm
-  oi := matchcontinue exp
+  oi := match exp
     local
       Real r;
       Integer i;
+      Option<Integer> op;
     case (DAE.RCONST(real = r))
       equation
         i = realInt(r);
-        true = realEq(r,intReal(i));
-      then SOME(i);
+        op = if realEq(r,intReal(i)) then SOME(i) else NONE();
+      then op;
     else NONE();
-  end matchcontinue;
+  end match;
 end realExpIntLit;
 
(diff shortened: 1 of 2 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2017-12-16_12-02-28](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5044/console).

sjoelund:
libgc uses sigprocmask... And third party simulation libraries might as well.

hkiel:
@sjoelund I wonder if meta_modelica_segv.c line 145 works as expected:
`sigprocmask()` is called with a successive call to `longjmp()` which should revert the change immediately!?

hkiel:
@sjoelund @casella I wonder why CombiTimeTable takes so much time on my machine (OSX) compared to libraries.openmodelica.org (Linux). When I run the N=1000 test the Frontend on my machine takes about 27s, whereas the hudson slave takes only 2.3s. For other tests I don't have a factor 10 in speed difference.

sjoelund:
It seems to be mostly frontend performance (I get slower numbers on my Linux desktop as well, although only a factor 2x or so; and my machine is usually a bit faster than the server). The new frontend should be faster on the frontend part though (and has several phases shown by execstat).

hkiel:
Which files of the frontend will be kept? Or is everything redesigned?

perost:
@hkiel Absyn/SCode is kept, otherwise pretty much everything is redesigned. The new frontend has it's own types instead of the ones declared in DAE.mo, so very little of the old code that can be reused as it is. Things like ExpressionSimplify would be fairly easy to rewrite for the new NFExpression type though, it's not that different to the old DAE.Expression.

PR #2092, opened 2018-01-03
loadFile(".../package.mo") ignores MODELICAPATH

sjoelund:
Note: If the same directory contains other libraries that are loaded
according to uses-annotations, the full MODELICAPATH is used for these.
This means you might end up loading only one of the packages from the
given directory.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-01-03_15-25-32](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5065/console).

PR #2105, opened 2018-01-12
Fixes Ticket:4618

bernhard-thiele:
Adds support for clocked when clauses in states.

### DIFF
--- a/Compiler/FrontEnd/StateMachineFlatten.mo
+++ b/Compiler/FrontEnd/StateMachineFlatten.mo
@@ -35,7 +35,7 @@
   description: Flattening of state machines
 
   This module contains functions to transform an instantiated state machine to flat data-flow equations.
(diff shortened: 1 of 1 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-01-12_11-49-30](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5087/console).

sjoelund:
I think the commit summary and contents are the wrong way around; should be "Adds support for clocked when clauses in states" as its main purpose (shows up in all summaries). The ticket number is a minor detail.

bernhard-thiele:
OK, I think I have seen this style of referencing the ticket in the summary quite a few times at some places. But I agree with your argument, it would be better to do it the other way round.

PR #2115, opened 2018-01-19
Avoid OMEdit crash when exporting a FMU on Windows

adeas31:


adeas31:
@adrpo There is no ticket about it. The model that user provided for testing is confidential.

adeas31:
I know but somehow this makes things work. We can run the testsuite and see. I will start hudson without push.

adeas31:
Well I started hudson and just realized that hudson runs the tests on Linux so it doesn't make any difference.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-01-19_15-48-56](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5114/console).

adrpo:
It could be that MinGW GCC generates bad code on Windows and this circumvents somehow the problem.

adeas31:
See #2141 and #2149.

PR #2120, opened 2018-01-23
[NF] Various fixes.

perost:
- Evaluate all constants in Typing.typeExp by default, and disable
  the package constant replacing phase since it's no longer needed.
- Don't evaluate iterators in Ceval.
- Use a type alias for ExpOrigin instead of Integer, to make it
  clearer what it is.
- Add generic error for unbound constants.

### DIFF
--- a/Compiler/FrontEnd/MetaModelicaBuiltin.mo
+++ b/Compiler/FrontEnd/MetaModelicaBuiltin.mo
@@ -260,7 +260,7 @@ function intBitRShift
   output Integer o;
 external "builtin";
 annotation(Documentation(info="<html>
-<p>Bit-wise right shift (like C, i << s).</p>
+<p>Bit-wise right shift (like C, i >> s).</p>
 </html>"));
 end intBitRShift;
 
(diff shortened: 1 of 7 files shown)


perost:
The error check for unbound constants will likely break a few models that previously passed the frontend, because we don't yet handle record constants where the bindings are on the record's fields instead of the constant itself. Hudson might call that a regression, but I think it's better with an error that indicates what's actually wrong rather than pretending to succeed.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-01-23_15-30-53](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5134/console).

casella:
Very nice [results](https://libraries.openmodelica.org/branches/history/newInst/2018-01-23%2015:08:34..2018-01-23%2016:56:16.html); ok to have "regressions" that actually reflect the status more realistically. 

26 models now simulate in [Modelica 3.2.2](https://libraries.openmodelica.org/branches/newInst/ModelicaTest_3.2.2/ModelicaTest_3.2.2.html), which is a big step forward compared to the few ones of recent times.

PR #2122, opened 2018-01-25
[NF] Propagate comments to the DAE.

perost:
- Add comments to Component and use ElementSource for
  equations/algorithms instead of SourceInfo.
- Generate comments for generated connect equations.
- Fix instantiation of if-statements so it doesn't add an else branch
  when there is none.

### DIFF
--- a/Compiler/FrontEnd/ElementSource.mo
+++ b/Compiler/FrontEnd/ElementSource.mo
@@ -412,6 +412,8 @@ algorithm
   end match;
 end getStatementSource;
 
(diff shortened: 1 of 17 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-01-25_15-45-57](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5140/console).

PR #2141, opened 2018-01-31
move some threadData vars to global ones

adrpo:
this will fix some issues for Windows:
- symboltable (OMEdit crash on startup)
- codegen function list (FMU generation from OMEdit)

### DIFF
--- a/Compiler/Global/Global.mo
+++ b/Compiler/Global/Global.mo
@@ -46,7 +46,6 @@ constant Integer instOnlyForcedFunctions = 0;
 constant Integer simulationData = 0 "For simulations";
 constant Integer codegenTryThrowIndex = 1;
 constant Integer codegenFunctionList = 2;
-constant Integer symbolTable = 3;
 
 // Global roots start at index=9
 constant Integer instHashIndex = 9;
@@ -64,6 +63,7 @@ constant Integer currentInstVar = 22;
 constant Integer operatorOverloadingCache = 23;
 constant Integer optionSimCode = 24;
 constant Integer interactiveCache = 25;
+constant Integer symbolTable = 27;
 
 // indexes in System.tick
 // ----------------------


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2018-01-31_16-03-15](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5177/).

hkiel:
Could this bug be seen in OMNotebook in v1.12 as well?

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-01-31_16-20-25](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5178/console).

adrpo:
@hkiel: this should fix all the QT clients on Windows for now, we will need to look into more details later why threadData local roots gets corrupted.

adrpo:
@hkiel: no version previous to 1.13 are affected, these were some recent changes from @sjoelund that changed how the SymbolTable is handled.

sjoelund:
Moving data to global data might break some clients because now you can't have two different symbol tables. Having them local is done for a very good reason...

adrpo:
@sjoelund: true, but there is something fishy with threadData->localRoots. They are not added as GC roots in Windows it seems and they are collected! 

Using `export GC_DONT_GC=true` OMEdit doesn't crash anymore. 

I even tried adding `GC_add_roots(&threadData->localRoots[0], &threadData->localRoots[MAX_LOCAL_ROOTS])` and it still doesn't help.

No idea what's going on here.

sjoelund:
@niklwors I suppose this affects you as well. Windows bugs relating to the garbage collection when trying to interface directly to OMC. You have had some similar weird bugs.

PR #2145, opened 2018-02-01
Fix memory corruption in ZeroMQ init

sjoelund:


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-02-01_08-08-38](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5184/console).

Compiler/runtime/zeromqimpl.c
```diff
@@ -71,22 +58,12 @@ void* ZeroMQ_initialize()
...
 #if defined(__MINGW32__) || defined(_MSC_VER)
-  if (zeroMQFileSuffix != NULL) {
-    zeroMQFilePath = (char*)malloc(strlen(tempPath) + strlen("/openmodelica.port.") + strlen(zeroMQFileSuffix) + 1);
-    sprintf(zeroMQFilePath, "%s/openmodelica.port.%s", tempPath, zeroMQFileSuffix);
-  } else {
-    zeroMQFilePath = (char*)malloc(strlen(tempPath) + strlen("/openmodelica.port") + 1);
-    sprintf(zeroMQFilePath, "%s/openmodelica.port", tempPath);
-  }
+  zeroMQFilePath = (char*)malloc(strlen(tempPath) + strlen("/openmodelica.port") + strlen(zeroMQFileSuffix) + 1);
+  sprintf(zeroMQFilePath, "%s/openmodelica.port%s", tempPath, zeroMQFileSuffix);
```
adeas31: This broke OMPython. You removed a `.` after port.

sjoelund: It doesn't break it. The suffix was fixed. If it was ".port", it still will be. If it was "port", it will not contain the dot.

adeas31: No it does. Previously it was that if there is no suffix then just use `openmodelica.port` and if suffix then `openmodelica.port.suffix`. You changed it to `openmodelica.portsuffix`. Fix this or update OMPython.

PR #2160, opened 2018-02-05
[NF] Implement proper handling of subscripts.

perost:
- Change typing of crefs so that subscripts aren't taken into account
  when typing each part, only for the whole cref expression.
- Implement/fix expansion of subscript slices.
- Fix type checking of array dimensions so that it actually gives an
  error message and fails on mismatched dimensions.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-02-05_15-24-56](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5206/console).

PR #2175, opened 2018-02-07
[NF] Better handling of constants.

perost:
- Take the subscripts into account when determining the variability
  of a cref, so we don't try to e.g. evaluate a constant with
  non-constant subscripts.
- Turn on collection of package constants again, since they might be
  used in the flat model due to the above change.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-02-07_16-46-55](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5224/console).

PR #2181, opened 2018-02-09
[C Runtime] added symbolical jacobian support to ida

wibraun:


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2018-02-09_14-49-50](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5230/).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-02-09_15-25-05](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5231/console).

casella:
I guess this should boost the performance significantly in case you have systems with Jacobians whose elements are mainly constants, because you just evaluate them at compile time (or at run-time if they depend on discrete variables) instead of computing the finite difference all the time.

Am I correct in this assumption?

wibraun:
Yes, such case should also be faster, but actually I have an other commit up one's sleeve which should boost the symbolic Jacobians significantly, because I can reuse of the matrix and the factorization of algebraic loops for every column. 
Therefore I launched the ScalableTestsuite with symbolic Jacobians on Hudson, let's see if we can see a significant improvement.

casella:
Excellent! BTW, most examples on the ScalableTestSuite have almost constant Jacobians.

PR #2186, opened 2018-02-12
[Backend] fix generateSymbolicJacobian config flag

wibraun:


lochel:
This shouldn't be merged. See https://github.com/OpenModelica/OpenModelica-testsuite/commit/fd4a590aadd21c4c3953e30b008abc9144475ad9#r27509076 for more information.

wibraun:
The referenced comment has no information about why this specific config flag is deprecated. 

1. I want to use it. 
2. it's not really redundant, since it inference actually also with the module detectJacobianSparsePattern what gets fixed by this commit. 
3. I think it makes sense to keep this config flag to distinguish in future different symbolic jacobians for different integration methods.

Please elaborate why you think this config flag should be removed and why you think it's okay to kill my hudson jobs.

lochel:
The flag `--generateSymbolicJacobian`, like the other deprecated flags, don’t play well with the opt. module scheduling. They interfere with the new flags, e.g. `--postOptModules+=`, and may lead to unexpected behaviour if combined. As I said before, they are deprecated and will be removed in the future. We kept them for the time being to make the transition easier for users.

wibraun:
Can explain what you mean by
> "don’t play well with the opt. module scheduling"
And what is the difference to the other similar flags, which are not deprecated and used without issues (e.g. partlintorn, maxSizeSolveLinearSystem).

Moreover, this config flag needs be replace by "--postOptModules-=detectJacobianSparsePattern --postOptModules+=generatedSymbolicJacobian", so I prefer just to use "--generatedSymbolicJacobian" also from the user perspective.

casella:
Gentlemen, you can't fight in here, this is the war room! [(_Stanley Kubrick, Dr. Strangelove_](https://www.google.it/url?sa=t&rct=j&q=&esrc=s&source=web&cd=1&cad=rja&uact=8&ved=0ahUKEwieqvi2jqPZAhWEUhQKHYXzDcEQ3ywIKDAA&url=https%3A%2F%2Fwww.youtube.com%2Fwatch%3Fv%3DUAeqVGP-GPM&usg=AOvVaw0n0wh4AHbtuLupW6lyKci4)) 😃 

Seriously, I understand there is a user requirement issue here, i.e., in order to carry out the back-end processing correctly in some cases, it only makes sense to switch on and off certain modules, which are not obvious to the end user but only to the developer. So, just removing `--generateSymbolicJacobian` and expecting that the end user knows the right combination to obtain the same effect is not a viable solution. On the other hand, we have to make sure there are no (or as little as possible) unwanted side effects. 

What if `--generatedSymbolicJacobian` is simply replaced by `--postOptModules-=detectJacobianSparsePattern --postOptModules+=generatedSymbolicJacobian` applied exactly in that place and in that order?

lochel:
That wouldn’t really solve the issue. I would propose to merge both modules and keep `--generatedSymbolicJacobian` to control the behaviour. The problem is just that ordinary flags should not mess up the module scheduling. I will do some clean-up and mark some of the new flags as deprecated.

casella:
I would suggest that you agree on a solution strategy before starting the PR

wibraun:
>I would propose to merge both modules and keep --generatedSymbolicJacobian to control the behaviour. 

That would work, so we can push this PR and remove the deprecated status and I'll merge this modules in a further commit.

lochel:
That doesn't make too much sense. Don't push something just to replace it with a next commit.

wibraun:
I think we agreed to keep --generatedSymbolicJacobian, so to remove the deprecated status will not replaced with the next commits, but the usage of this flag will be extended.

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2018-02-16_16-43-35](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5294/).

lochel:
Again, this change is bad and should not be pushed. Merge both postOptModules (detectJacobianSparsePattern and generatedSymbolicJacobian) and use the flag `--generatedSymbolicJacobian`  to control the behaviour.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-02-16_17-19-17](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5296/console).

PR #2191, opened 2018-02-13
Add a warning for StateSelect.always/never

sjoelund:
If a non-state variable has StateSelect.always or a state variable has
StateSelect.never, we now give a warning during the SimCode phase.

This resolves ticket:3689.

### DIFF
--- a/Compiler/FrontEnd/ComponentReference.mo
+++ b/Compiler/FrontEnd/ComponentReference.mo
@@ -1271,6 +1271,16 @@ algorithm
   end match;
 end isPreCref;
 
+public function isPreviousCref
+  input DAE.ComponentRef cr;
+  output Boolean b;
+algorithm
+  b := match(cr)
+    case(DAE.CREF_QUAL(ident=DAE.previousNamePrefix)) then true;
+    else false;
+  end match;
+end isPreviousCref;
+
 public function isStartCref
   input DAE.ComponentRef cr;
   output Boolean b;
--- a/Compiler/SimCode/SimCodeUtil.mo
+++ b/Compiler/SimCode/SimCodeUtil.mo
@@ -8292,19 +8292,19 @@ algorithm
     SimCode.SparsityPattern sparsityT;
   case(SOME(dmd)) algorithm
     print("\ndaeMode: \n" + UNDERLINE + "\n");
-	  str := "residual Equations:\n"+UNDERLINE+"\n";
-	  print(str);
-	  dumpSimEqSystemLst(List.flatten(dmd.daeEquations),"\n");
-	  dumpVarLst(dmd.residualVars,"residualVars("+intString(listLength(dmd.residualVars))+")");
-	  dumpVarLst(dmd.algebraicVars,"algebraicDAEVars("+intString(listLength(dmd.algebraicVars))+")");
-	  dumpVarLst(dmd.auxiliaryVars,"auxVars("+intString(listLength(dmd.auxiliaryVars))+")");
-	  if isSome(dmd.sparsityPattern) then
-	    str := "Sparsity Pattern:\n"+UNDERLINE+"\n";
-	    print(str);
-	    SimCode.JAC_MATRIX(sparsityT=sparsityT) := Util.getOption(dmd.sparsityPattern);
-	    dumpSparsePatternInt(sparsityT);
-	  end if;
-	then ();
+    str := "residual Equations:\n"+UNDERLINE+"\n";
+    print(str);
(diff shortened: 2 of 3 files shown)


OpenModelica-Hudson:
The job failed the coding style; for details, see [OpenModelica_TEST_PULL_REQUEST 2018-02-13_11-37-42](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5249/console). Remember to clean up the commit you did that breaks the coding style rather than piling a new commit on top of the old ones. This goes double for accidental addition of files that you need to remove.

casella:
I'm not sure about the rationale of this PR. Consider this model

    model Tank
      Real M(stateSelect = StateSelect.never);
      Real y(stateSelect = StateSelect.always);
      parameter Real A = 1;
      parameter Real rho = 1000;
      Real w = if time < 1 then 0 else 1;
    equation
      der(M) = w; 
      M = rho*A*y;
    initial equation
      y = 1;
    end Tank;

This seems to me a perfectly legitimate case of state variable change, though of course one could get the new state variable set by using prefer and avoid instead. Why should this generate warnings?

sjoelund:
Always means always. Anyway, we let the model pass with a warning since you might get unexpected behavior. Indeed you should use prefer if you would prefer it to be a state but it doesn't have to be a state. For things like reinit or if you want to control a variable through interactive simulation it's good to be able to get a warning if the compiler couldn't select a state for you (for reinit I guess it's even an error if it fails).

casella:
All right. In fact, I always discourage use of never and always, except for debugging purposes.

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2018-02-28_09-43-23](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5361/).

sjoelund:
Modelica.Mechanics.MultiBody.Examples.Elementary.RollingWheel has some problems with bad stateSelect attributes. It sets StateSelect.always to all variables in x, y, angles and der_angles. But only 1 of 3 der_angle variables is possible to select as a state. Changing the global setting in there to StateSelect.prefer makes OM fail for dummy derivatives. The following does work and silences the warning, but I'm a bit unsure if it's the correct approach for MSL since other tools could select different states?
```diff
model RollingWheel
  "Single wheel rolling on ground starting from an initial speed"
   extends Modelica.Icons.Example;

  Modelica.Mechanics.MultiBody.Parts.RollingWheel wheel1(
    wheelRadius=0.3,
    wheelMass=2,
    wheel_I_axis=0.06,
    wheel_I_long=0.12,
    hollowFraction=0.6,
    x(start=0.2),
    y(start=0.2),
-    der_angles(start={0,5,1})
+    der_angles(start={0,5,1},stateSelect={StateSelect.prefer,StateSelect.prefer,StateSelect.always}))
```

Or the following in Parts.mo:
```diff
    SI.AngularVelocity der_angles[3](
      start={0,0,0},
      each fixed=true,
-      each stateSelect=stateSelect) 
+      each stateSelect=if stateSelect==StateSelect.always then StateSelect.prefer else stateSelect) "Derivative of angles"
      annotation (Dialog(group="Initialization",showStartAttribute=true));
```
The problem in RevoluteConstraint ([Modelica 3.2.1/Mechanics/MultiBody/Joints.mo:260:5-270:3:writable] Warning: Variable joint.phi has attribute stateSelect=StateSelect.never, but was selected as a state) seems to be that StateSelect.never is used, but it is used for both phi and w and we always have der(w) in the model, so that seems like a really bad idea.

@casella @tobolar Do you think this should be resolved in MSL?

casella:
Absolutely, and we have a good chance of doing it for the next maintenance revision. Please open a PR on the MSL repo.

tobolar:
Well the state selection is not done properly in the RollingWheel model and it is necessary to implement a plausible model. So an issue in MSL is desirable.

tobolar:
@casella 
>  I always discourage use of never and always

I'm not satisfied with that strategy. In multi-body models, the joints' states are per default set to prefer. But for models with kinematc loops, it is often wise to explicitly set one (or more) of the variables to always be state.
The strategy where that variable should be prefered prior to always would mean that the other variables have to be selected as default states in order to make clear suggestion on state selection. 
I guess at least for multi-body systems that strategy is not well feasible.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-03-05_11-37-16](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5375/console).

PR #2211, opened 2018-02-16
[BE] Do not run clockPartitioning unless needed

sjoelund:
This resolves ticket:4756 by keeping track of when clocked operators
have been used.

### DIFF
--- a/Compiler/BackEnd/SynchronousFeatures.mo
+++ b/Compiler/BackEnd/SynchronousFeatures.mo
@@ -37,27 +37,29 @@ encapsulated package SynchronousFeatures
                - sub-clock partitioning"
 
 
-public import Absyn;
-public import BackendDAE;
-public import DAE;
+import Absyn;
+import BackendDAE;
+import DAE;
 
-protected import BackendDAEOptimize;
-protected import BackendDAEUtil;
-protected import BackendDump;
-protected import ExpressionDump;
-protected import BackendEquation;
-protected import BackendVariable;
-protected import ComponentReference;
-protected import DAEUtil;
-protected import DAEDump;
-protected import Error;
-protected import Flags;
-protected import List;
-protected import Util;
-protected import Types;
-protected import Expression;
-protected import HashTable;
-protected import MMath;
+protected
+import BackendDAEOptimize;
+import BackendDAEUtil;
+import BackendDump;
+import ExpressionDump;
+import BackendEquation;
+import BackendVariable;
+import ComponentReference;
+import DAEUtil;
+import DAEDump;
+import Error;
+import Expression;
+import Flags;
+import Global;
+import HashTable;
+import List;
+import MMath;
+import Types;
+import Util;
 
 // =============================================================================
 // clock partitioning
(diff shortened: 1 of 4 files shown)


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2018-02-16_11-19-44](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5291/).

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2018-02-21_11-25-51](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5314/).

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2018-02-21_12-25-29](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5316/).

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2018-02-21_13-05-33](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5318/).

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2018-02-21_13-23-54](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5319/).

casella:
It really seems like clockPartioning is doing more than just finding out clock partitions. Any chance to get a good understanding of what that actually is?

sjoelund:
I don't know. I just ran this job a bunch of times while testing VisualStudio-related things

lochel:
The module is also calculating independent partitions to run other modules with smaller equation sets.

casella:
@lochel: OK. So, that will mean that if there are functions in the backend whose complexity is (significantly) greater than O(N), the overall time should be smaller if there are such independent systems, as long as this is not offset by the time taken by this module.  Are you aware of any such functions, besides those who scale badly due to sub-optimal implementation (see [#4146](https://trac.openmodelica.org/OpenModelica/ticket/4146))?

Anyway, I would suggest to change its name to systemPartitioning, as it doesn't necessarily involve clocks. Unless the system partitioning and the clock partitioning can be split in two separate modules, is that the case?

Sorry for asking about all these details, but all the documentation of this module says is 

    clockPartitioning (Does the clock partitioning.)

which is incorrect and a bit misleading. In fact, there was an old flag (disablePartitioning) which was more correctly described as

    Deactivates partitioning of entire equation system. 

Unfortunately this is now deprecated in favour of clockPartitioning

lochel:
@casella No, I am not aware of further modules. But that doesn't mean too much...

I think it is reasonable to rename that module. I am fine with the proposal `systemPartitioning`.

@sjoelund If you want, I can take over this PR and implement the switch to deactivate the clock partitioning.

sjoelund:
@lochel Sure, you can push to my branch I'm sure. Then we can rename it and see if it helps performance to disable only the clock partitioning part.

PR #2216, opened 2018-02-16
fix crashes in omccAPI if threadData is allocated in the heap

adrpo:
- hide threadData again
- set threadData and then copy it after each call


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-02-16_22-49-06](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5299/console).

PR #2217, opened 2018-02-16
Merge postOptModules that detect sparse pattern

lochel:
Merge postOptModules detectJacobianSparsePattern and generateSymbolicJacobian that detect sparse pattern.

Flag `--generateSymbolicJacobian` can be used to additionally generate the symbolic Jacobian itself.


lochel:
The new postOptModule is called `symbolicJacobian` for now. Please feel free to make better suggestions.

@wibraun Please note: This is now actually what #2186 was supposed to do. I keep ticket:4769 untouched, since it is about enhancing the current implementation and this PR is about fixing interferences with the opt. module pipeline.

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2018-02-17_00-32-18](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5300/).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-02-17_00-54-34](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5301/console).

wibraun:
Sorry, I've still no idea why it is so bad to enable and disable modules, please help me to understand, which issues are raised by such behaviour.

Okay, obviously is it better to reduce the amount of modules, if possible, however as far as I understood this is not the central issue here.

wibraun:
@lochel I'm asking that, because I'm wondering if we have also issues with other config flags, which enables or disables optimization modules (see [BackendDAEUtil.getPostOptModulesString](https://github.com/OpenModelica/OMCompiler/blob/master/Compiler/BackEnd/BackendDAEUtil.mo#L7771)  ).

lochel:
It is the same with the other flags that try to cooperate with the module pipelines. Config flags can be used to adjust the behaviour of modules, but shouldn't manipulate the module scheduling.

From a user’s point of view, it would also be better to have less flags that magically interact with others. Ideally, there would also be a short section for each module in the user’s guide.

PR #2218, opened 2018-02-19
Improve reporting memory usage

sjoelund:
Disabled Serializer.mo and wrote a dedicated memory usage counter. It
also tries to account for GC overhead (allocating 24 bytes actually
results in 32 bytes being allocated).

Added -d=execstatGCcollect, which adds a GC.gcollect() and an extra
notification for each execstat event so we can track how much memory is
in use for each phase. This has a roughly 3x performance penalty as full
GC is expensive. -d=reportSerializedSize is even slower though, and
together they cause a 10x hit to performance.

### DIFF
--- a/Compiler/SimCode/SimCodeMain.mo
+++ b/Compiler/SimCode/SimCodeMain.mo
@@ -94,7 +94,6 @@ import HpcOmTaskGraph;
 import SerializeModelInfo;
 import TaskSystemDump;
 import SerializeInitXML;
-import Serializer;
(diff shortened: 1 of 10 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-02-19_08-31-47](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5302/console).

PR #2220, opened 2018-02-19
Use set instead of unordered_set

sjoelund:
OSX has some difficulty using unordered_set so use set instead.
Performance is not important here since this is only used for debugging.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-02-19_13-19-00](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5305/console).

PR #2221, opened 2018-02-19
Remove the extension

adeas31:


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-02-19_15-24-38](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5306/console).

Compiler/Script/CevalScript.mo
```diff
@@ -1393,7 +1393,7 @@ algorithm
           if (0 == System.systemCall("unzip -q -o -d \"" + workdir + "\" \"" +  filename + "\"")) then
             b = true;
             s1 = System.basename(filename);
-            s2::_ = System.strtok(s1, ".");
+            s2 = Util.removeLast3Char(s1);
```
hkiel: Is it safe to remove a fixed amount of chars from the end???

adeas31: Yeah since I expect the filename with a `.mol` extension but I think its still a good idea to check the length of the string.

PR #2228, opened 2018-02-23
Improve report of used size of MM datatypes

sjoelund:
We now collect both how many bytes we want to allocate and how much
memory libGC will consume and report this when reportSerializedSize
is used.

Also fixed a bug where the incorrect size was calculated for the GC
overhead: GC uses 1 byte internally so a 15-byte allocation allocates
a single 16-byte granule whereas a 16-byte allocation allocates 2 16-
byte granules. This accounts for a 20% overhead that was not accounted
for previously.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-02-23_09-43-14](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5328/console).

PR #2235, opened 2018-02-26
[NF] Operator overloading support.

mahge:
-  Initial operator overloading support for the new front-end.

  - Some error messaages and extra checks needs improvments.
  - A few other minor modifications.
    - new type-matching type PLUG_COMAPTIBLE.
      - used when matching is possible on a component basis.
    - use different matching rules for different cases.
      - e.g matchings allowed for assignments are different from matchings allowed for arguments.
      - needs more improvment.
    - Type 'String' calls as special builtin.
    - ...


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-02-26_10-38-41](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5337/console).

PR #2246, opened 2018-03-01
Enable Cpp logger under Windows and finalize it, ticket:4773

rfranke:
The finalizer shall flush buffers, e.g. of LoggerXMLTCP.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-03-01_22-51-37](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5367/console).

SimulationRuntime/cpp/Makefile.omdev.mingw
```diff
@@ -81,7 +81,7 @@ else
 	PARALLEL_OUTPUT_COMMAND=-DUSE_PARALLEL_OUTPUT=OFF
 endif
 
-USE_LOGGER="false"
+USE_LOGGER="true"
 ifeq ("$(USE_LOGGER)","true")
```
adeas31: Unfortunately this will not work. This will evetually become,
`ifeq (""true"","true")`
and will fail. Either get rid of quotes from the assignment or from the if condition. See also my comment on ticket:4773.

adrpo: The quotes are removed with: https://github.com/OpenModelica/OMCompiler/pull/2252

PR #2266, opened 2018-03-12
[NF][#4784] Fix conversion of complex types to old DAE.

mahge:
- Create proper DAE.TYPES_VAR for complex types.
  - create DAE complex equation for complex equations
    - until we start scalarizing complex equations.

### DIFF
--- a/Compiler/NFFrontEnd/NFBinding.mo
+++ b/Compiler/NFFrontEnd/NFBinding.mo
@@ -250,5 +250,25 @@ public
     end match;
   end isEqual;
 
+  function toDAE
+    input Binding b;
+    output DAE.Binding outb;
+  algorithm
+    outb := match b
+      case UNBOUND() then DAE.UNBOUND();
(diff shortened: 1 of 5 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-03-12_15-27-36](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5400/console).

mahge:
@casella Yes. It is not completely fixed yet. There are still connection and general equation issues. I don't think it is going to be an easy fix. So far I am tracking down what exactly generates these extra equations. There is very little done concerning connectors with record components. I will update you as I go.

casella:
OK. 

Maybe @perost can help you with connection equations involving records. BTW, if I am not mistaken, these are also required for MultiBody connectors, so it's not just a problem with Complex numbers.

PR #2267, opened 2018-03-12
[NF] Fix Expression.mapShallow.

perost:
- Fixed the reference checks in mapShallow, that previously were
  reversed leading to changed subexpressions being thrown away.
  This caused e.g. the evaluation of end to not have any effect
  if the end was part of a larger expression.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-03-12_15-07-03](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5399/console).

PR #2276, opened 2018-03-15
[BE] differentiate the inner exp of getPart

wibraun:


Compiler/BackEnd/Differentiate.mo
```diff
@@ -1462,7 +1462,10 @@ algorithm
       exp_1 = Expression.makeZeroExpression(Expression.arrayDimension(tp));
     then (exp_1, inFuncs);
 
-    case ("$getPart",_) then (exp, inFuncs);
+    case ("$getPart",_) equation
+        (exp_1, funcs) = differentiateExp(exp, inDiffwrtCref, inInputData,inDiffType,inFuncs, maxIter);
+    then (exp_1, inFuncs);
```
lochel: What is `$getPart` and why aren't you returning the updated function tree?

wibraun: Oh, good point. Of course it has to be the new function tree.

As far as I understood @rfranke `$getPart` is a transition between a continuous variable and the discrete variable an an internal OpenModelica representation. The source of it, is one of the commit inside this squash: [OMCompiler/db90b5d5](https://github.com/OpenModelica/OMCompiler/commit/db90b5d56615e9487df1890824b40c9160576ff7)

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2018-03-15_17-49-20](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5418/).

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2018-03-15_18-19-17](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5420/).

wibraun:
@sjoelund In the commit of Per the 3rdParty has been downgraded, by an oversight, I guess, but now nothing can't be pushed, since an update of 3rdParty submodule is used in the next run, for some reason. Any suggestion how to solve this?

lochel:
I've updated the 3rdParty subproject for you.

What models are affected by this change? It is still not clear to me what these expressions represent.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-03-15_22-35-52](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5427/console).

wibraun:
I've also tried to update 3rdParty subproject, but it was not pass the testsuite. Don't know why it pass in your pass for you. The question is why it was possible to downgrade 3rdParty .

lochel:
@rfranke, can you help us here. It seems to be unclear what `getPart` actually is and why it needs to be differentiated.

wibraun:
It is introduced in [SynchronousFeatures.mo](https://github.com/OpenModelica/OMCompiler/blob/master/Compiler/BackEnd/SynchronousFeatures.mo#L2123) and it is used to get the related discrete partition, if a continuous variable is in connected to a discrete one.

It's not the `$getPart` operator that needs to be differentiated, but the expression that is inside, what is needed because the jacobian is created under the assumption, that all partitions are activated. 
It might be helpful to have a look at the [DiscreteExample](https://github.com/OpenModelica/OpenModelica-testsuite/blob/master/openmodelica/fmi/ModelExchange/2.0/testDiscreteStructe.mos), where the Jacobian is plainly wrong if the inner expression of the `$getPart` is not differentiated.

PR #2277, opened 2018-03-15
Forbid constant values in match inputs.

perost:


sjoelund:
I would just make it a notification. Just like if true is useful, this might be as well.

perost:
@sjoelund Notifications are easy to overlook, but fair enough I guess. I'll just make it a notification then and push it in.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-03-15_15-54-07](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5416/console).

PR #2293, opened 2018-03-20
[NF] Redeclare and extends fixes.

perost:
- Avoid expanding shared class extends nodes before cloning them.
- Added check for short class definition self reference (class A = A).


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-03-20_13-07-26](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5446/console).

PR #2303, opened 2018-03-21
[NF] Fix issue with converting functions to DAE.

perost:
- Convert the function tree before everything else, to avoid function
  nodes being clobbered if the function is used as a type somewhere.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-03-21_20-26-54](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5462/console).

PR #2307, opened 2018-03-23
[NF] Record binding improvements.

perost:
- Evaluate record constructors to record expressions.
- Split record expressions when flattening a complex component.
- Added dummy node to the definition of the Integer function instead
  of using an empty node, so we can assume each function has a node.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-03-23_13-02-08](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5472/console).

PR #2311, opened 2018-03-24
improved inInline.getFunctionInputsOutputBody

vruge:
use loop instead of recursive match

### DIFF
--- a/Compiler/FrontEnd/Inline.mo
+++ b/Compiler/FrontEnd/Inline.mo
@@ -854,7 +854,7 @@ algorithm
           (newExp1,assrtLst) = Expression.traverseExpBottomUp(newExp,function inlineCall(fns=fns),assrtLst);
         else // normal Modelica
           // get inputs, body and output
-          (crefs,lst_cr,stmts,repl) = getFunctionInputsOutputBody(fn,{},{},{},repl);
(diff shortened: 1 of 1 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-03-24_09-12-24](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5479/console).

Compiler/FrontEnd/Inline.mo
```diff
@@ -1110,71 +1110,71 @@ end addTplAssignToRepl;
...
       then fail();
-  end match;
+    end match;
+
+
+  end for;
+
+  oInputs := listReverse(oInputs);
+  oOutputs := listReverse(oOutputs);
+
```
hkiel: Is it necessary to reverse the list? you used listAppend() before, so I guess the order is not the same as the old version anyway...

vruge: The order should be importend, because
` newExp = Expression.makeTuple(list( VarTransform.getReplacement(repl,cr) for cr in lst_cr)); // lst_cr = oOutputs
`
will transform
`(a, b) = f(c) -> a=f(c)[2], b = f(c)[1]
`
, but maybe I wrong..

PR #2319, opened 2018-03-26
improved inStream

vruge:
use positiveMax like proposed in
see. https://trac.openmodelica.org/OpenModelica/ticket/3885#comment:12

refs ticket:3885, ticket:4441

### DIFF
--- a/Compiler/BackEnd/BackendDAEOptimize.mo
+++ b/Compiler/BackEnd/BackendDAEOptimize.mo
@@ -111,6 +111,122 @@ algorithm
 end simplifyAllExpressions;
 
 // =============================================================================
+// simplifyInStream
+//
+// OM introduces $OMC$PosetiveMax which can simplified using min or max attribute
+// see Modelica spec for inStream
+// author: Vitalij Ruge
+// see. #3885, 4441
+// =============================================================================
+
+public function simplifyInStream
+  input output BackendDAE.BackendDAE dae;
+protected
+  BackendDAE.Shared shared = dae.shared;
+  BackendDAE.EqSystems eqs = dae.eqs;
+  list<BackendDAE.Variables> vars = list(eq.orderedVars for eq in eqs);
+algorithm
+  // from the description inputs are part of globalKnownVars and localKnownVars :(
+  vars := shared.globalKnownVars :: vars; // need inputs with min = 0 or max = 0
+  vars := shared.localKnownVars :: vars;
+  _ := BackendDAEUtil.traverseBackendDAEExpsNoCopyWithUpdate(dae, simplifyInStreamWork, vars);
+end simplifyInStream;
+
+
+protected function simplifyInStreamWork
+  input DAE.Exp inExp;
+  input list<BackendDAE.Variables> inVars;
+  output DAE.Exp outExp;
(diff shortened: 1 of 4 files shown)


lochel:
Please fix the indentation in your changes.

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2018-03-26_22-37-02](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5492/).

vruge:
Unfortunately, I have no time for it

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-03-27_00-13-25](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5493/console).

Compiler/BackEnd/BackendDAEOptimize.mo
```diff
@@ -110,6 +110,122 @@ algorithm
...
+            case DAE.CALL(path=Absyn.IDENT("$OMC$PositiveMax"),expLst={e as DAE.UNARY(DAE.UMINUS(tp), DAE.CREF(componentRef=cr)), expr})
+              algorithm
+               (eMin, eMax) := simplifyInStreamWorkExpresion(cr, outVars);
+               isZero := simplifyInStreamWorkSimplify(eMax, true);
+               then
+                  if isZero then Expression.createZeroExpression(tp) else Expression.makePureBuiltinCall("max", {e, expr}, tp);
+
+            case DAE.CALL(path=Absyn.IDENT("$OMC$PositiveMax"),expLst={e, expr})
+               guard Expression.isZero(e)
+               then e;
```
sjoelund: This seems totally wrong. If e==0 you should probably return eps, right?

vruge: Unclear:
 1)  `0.max=0.min = 0` or `0.nominal = 0 //  eps = nominal?`  
 2) https://trac.openmodelica.org/OpenModelica/ticket/4441 is valid 
   - accept the effect of elimination  

if 1) and 2) are correct then 0 is correct. Otherwise I would accept eps, too

sjoelund: The 3.4 specification does not use the min or max attributes of the expression anywhere; only the actual value of the expression. And since a tool is free to merge min/max attributes (which is especially done for aliases), you should just assume for a literal value, if it is within the min/max bounds you can tighten the bounds anyway.

vruge: In this case  (eMin, eMax) defined as (SOME(e), SOME(e)) and we are in the same cases as about, right?

vruge: the case 0/0 is missing, maybe we can keep eps ... for the moment

vruge: if it's induce some bugs

casella:
See the comments in [#3885](https://trac.openmodelica.org/OpenModelica/ticket/3885#comment:14)

PR #2326, opened 2018-03-28
[NF] Cleaning and various optimizations.

perost:
- Cleaned up the name lookup a bit, and got rid of the error spamming
  from Function.lookupFunctionSilent.
- Changed the handling of Integer and String so that e.g. the type
  Integer and the function Integer are different entities, instead of
  e.g. using the same InstNode for both.
- Optimize the constants in NFBuiltin and NFBuiltinFuncs so that they
  become actual constants in the generated code, and not created each
  time they're used.
- Remove special handling of StateSelect and ExternalObject from the
  name lookup, since Modelica 3.4 clarifies that they aren't reserved
  names, and added ExternalObject to ModelicaBuiltin instead
  (StateSelect was already defined there).
- Implemented a better check for subscripts on elements that shouldn't
  be subscripted.

### DIFF
--- a/Compiler/NFFrontEnd/NFBuiltin.mo
+++ b/Compiler/NFFrontEnd/NFBuiltin.mo
@@ -46,7 +46,7 @@ import SCode;
 import Binding = NFBinding;
 import BindingOrigin = NFBindingOrigin;
 import NFClass.Class;
-import NFClassTree;
+import NFClassTree.ClassTree;
 import NFComponent.Component;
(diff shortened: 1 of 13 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-03-28_16-32-11](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5510/console).

PR #2328, opened 2018-03-28
[NF] Fix typing of calls with multiple outputs.

perost:
- Fix typing of function calls with multiple outputs so that only the
  first output is used when the call isn't alone on the rhs.
- Use correct origin in Typing.typeSubscripts.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-03-28_17-54-56](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5512/console).

PR #2332, opened 2018-03-29
added sum to skipped case for cse

vruge:
refs ticket:4850


lochel:
Why does the cse module fail for `sum`?

wibraun:
Thanks Vitalij, I think also it's a suitable fix as long as the function return type is not supported see (https://trac.openmodelica.org/OpenModelica/ticket/4826 ticket:4826).

PR #2341, opened 2018-04-05
Set equation kind of simple shared equations

lochel:


lochel:
@sjoelund Are you going to adapt the code generation?

sjoelund:
I don't have time for that. Note that if you have a getEquationAttributes function you can save a lot of code since you can do most of the work before the match-expression.

lochel:
It would require the same kind of match pattern to extract the equation attributes.

sjoelund:
Yes, but if you already have such a function the code is more compact and readable. Also more maintainable since you don't have copy-pasta.

wibraun:
Also the function is already there `BackendEquation.getEquationAttributes`.

vruge:
Equation with inputs should  have the kind=dynamic and not initial right?
(Inputs are part of globalKnows, too)

PR #2357, opened 2018-04-11
[BE] Added usage of failtrace debug flag

kabdelhak:
- BackendEquation.solveEquation

### DIFF
--- a/Compiler/BackEnd/BackendEquation.mo
+++ b/Compiler/BackEnd/BackendEquation.mo
@@ -2234,6 +2234,7 @@ algorithm
     then eq;
 */
     else equation
+      true = Flags.isSet(Flags.FAILTRACE);
       BackendDump.dumpBackendDAEEqnList({eqn}, "function BackendEquation.solveEquation failed w.r.t " + ExpressionDump.printExpStr(crefExp), true);
       Error.addInternalError("function solveEquation failed", sourceInfo());
     then fail();


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-04-11_18-16-17](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5557/console).

lochel:
Welcome @kabdelhak. Thanks for your first contribution.
Why do you want to hide the error message? Is the function supposed to fail and the test suite doesn't cover that case?

kabdelhak:
Thank you,

the commit was manly for me to get to used to the whole GIT-process. I stumpled upon it and asked Willi Braun about 
what that might be and we figured, that the function is allowed to fail and is not a crucial error. We thought it should just be shown 
if explicitly asked for. 

Am 08.05.18 12:55 schrieb Lennart Ochel  <notifications@github.com>:

lochel:
@kabdelhak, @wibraun Do you have a model where this function actually fails? It should obviously fail for algorithms, when equations, and if equations. However, if it is not expected to reach one of these cases, then I would rather keep the error message explicit.

wibraun:
@lochel It depends on the usage of this function. If one would like to use it as a try and in failing case do something else, than the error message is totally misleading, especially for end-users. So I think it's okay to hide it. In the new removeSimpleEquations module is used like that and therefore I proposed to hide the error messages instead of having them in all related tests.

PR #2358, opened 2018-04-11
some minor optimizations

hkiel:


Compiler/BackEnd/BackendDAEOptimize.mo
```diff
@@ -2014,24 +2014,20 @@ algorithm
...
-    case (e as DAE.CALL(path=Absyn.IDENT(name=opName)),_,(i1,i2,i3,i4,i5,i6,i7,i8)) equation
-      true = stringEq(opName,"sin") or stringEq(opName,"cos") or stringEq(opName,"tan");
+    case (e as DAE.CALL(path=Absyn.IDENT(name=opName)),_,(i1,i2,i3,i4,i5,i6,i7,i8))
+      guard stringEq(opName,"sin") or stringEq(opName,"cos") or stringEq(opName,"tan")
       then (e, (i1,i2,i3,i4+1,i5,i6,i7,i8));
 
-    case (e as DAE.CALL(path=Absyn.IDENT(name=opName)),_,(i1,i2,i3,i4,i5,i6,i7,i8)) equation
-      true = stringEq(opName,"der");
+    case (e as DAE.CALL(path=Absyn.IDENT(name="der")),_,(i1,i2,i3,i4,i5,i6,i7,i8))
       then (e, (i1,i2,i3,i4,i5,i6,i7,i8));
```
hkiel: @vwaurich @sjoelund Is it correct that no counter is increased here???
If so, then the tuple needs not to be recreated, otherwise a +1 is missing somewhere...

sjoelund: As far as I know, the `der` operator is expanded by the backend so it should be the same as a CREF (which is ignored in this code).

PR #2370, opened 2018-04-16
[NF] Improvements.

perost:
- Simplified the type checking of bindings, and made it work for
  binding on array types too.
- Added check that non-type classes shouldn't have dimensions.
- Fixed the check for 'each' being used correctly, and made it into
  a varning instead of an error.
- Fixed the typing of crefs so that it doesn't cause typing cycles.

### DIFF
--- a/Compiler/NFFrontEnd/NFBinding.mo
+++ b/Compiler/NFFrontEnd/NFBinding.mo
@@ -43,14 +43,15 @@ protected
   import Binding = NFBinding;
 
 public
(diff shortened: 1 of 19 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-04-16_18-36-03](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5579/console).

PR #2392, opened 2018-04-24
[NF] Improve subscripting of expressions.

perost:
- Handle typenames, ranges and reductions in
  Expression.applySubscriptIndex.
- Use Expression.applySubscriptIndex in Expression.expandGeneric,
  instead of just creating a subscripted expression.
- Fixed mistake in typing of Boolean ranges that resulted in them
  always getting a size of 1.
- Added case to Typing.typeIterator for when the iterator has already
  been typed (which can happen e.g. if the iterator is needed to
  determine the size of a dimension).

### DIFF
--- a/Compiler/NFFrontEnd/NFExpression.mo
+++ b/Compiler/NFFrontEnd/NFExpression.mo
@@ -53,6 +53,8 @@ public
   import Type = NFType;
   import ComponentRef = NFComponentRef;
   import NFCall.Call;
+  import Binding = NFBinding;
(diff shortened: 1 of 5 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-04-24_12-45-26](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5617/console).

PR #2396, opened 2018-04-24
Disable incorrect typing of .+ for Strings.

perost:
- Concatenation of a String scalar and array was previously defined
  incorrectly in OperatorOverloading, such that it would expect the
  scalar to be a Boolean instead of String. This fixes that particular
  issue, but the frontend still handles it wrong since it assumes that
  the .+ operator is commutative (which isn't true for Strings). This
  commit disables the incorrect Boolean matches, but keeps the old
  behaviour of String not working until someone fixes it properly or
  it's made obsolete by the new frontend.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-04-24_16-39-13](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5623/console).

PR #2402, opened 2018-05-02
Fix undefined behaviour in codegen of array reduction

sjoelund:
There was a condition variable that was never assigned to that caused
the C-compiler to either assume the reduction always failed, always
succeeded or at runtime get random memory that made the choice.
The variable and superfluous check have been removed.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-05-02_08-12-08](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5638/console).

PR #2411, opened 2018-05-05
[NF] Various improvements.

perost:
- Add manual typing of cardinality and Connections.*.
- Fixed typing of if equations when the condition contains one of the
  above mentioned connection operators.
- Added evaluation of OpenModelica_uriToFilename.
- Improved Expression.applyIndexSubscript for crefs.

### DIFF
--- a/Compiler/NFFrontEnd/NFCall.mo
+++ b/Compiler/NFFrontEnd/NFCall.mo
@@ -389,6 +389,7 @@ uniontype Call
 
     (callExp, ty, variability) := match ComponentRef.firstName(cref)
       case "String" then typeStringCall(call, origin, info);
+      case "branch" then typeBranchCall(call, origin, info);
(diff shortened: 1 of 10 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-05-05_15-35-15](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5651/console).

casella:
1421 improvements on the [testsuite](https://libraries.openmodelica.org/branches/history/newInst/2018-05-04%2021:07:03..2018-05-05%2016:33:27.html) is really an impressive result. Too bad there is only one more model that actually simulates, but anyway we're making very good progress.

perost:
@casella: It's because this only fixes the typing of the connection operators, but they also need to be evaluated by the connection handling to make the models work (i.e. handling of cardinality and overconstrained connections).

PR #2414, opened 2018-05-07
[NF] Function evaluation improvements for records.

perost:
- Build a record expression from the child nodes if a record output
  doesn't have an explicit binding.
- Evaluate expressions in function outputs and local variables too
  (currently in order of declaration, needs to be dependency sorted).
- Added check that the outputs are assigned a value when evaluation a
  function (by adding Expression.EMPTY() and using it as the initial
  value if an output has no binding).
- Always evaluate complex bindings on parameters, even if they aren't
  structural, since they need to be split into their component values.

### DIFF
--- a/Compiler/NFFrontEnd/NFBinding.mo
+++ b/Compiler/NFFrontEnd/NFBinding.mo
@@ -156,6 +156,12 @@ public
           binding.bindingExp := exp;
         then
           ();
+
+      case FLAT_BINDING()
+        algorithm
+          binding.bindingExp := exp;
+        then
+          ();
     end match;
   end setTypedExp;
 
--- a/Compiler/NFFrontEnd/NFEvalFunction.mo
+++ b/Compiler/NFFrontEnd/NFEvalFunction.mo
@@ -42,6 +42,7 @@ import Binding = NFBinding;
 import NFComponent.Component;
 import Type = NFType;
 import Dimension = NFDimension;
+import NFClassTree.ClassTree;
 
 protected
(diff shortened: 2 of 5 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-05-07_16-24-40](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5654/console).

perost:
Note that regressions are expected due to these changes, since the function evaluation can't handle everything yet. I pushed it in anyway to get a better idea of what needs to be fixed.

casella:
Indeed 😄 
(see [report ](https://libraries.openmodelica.org/branches/history/newInst/2018-05-07%2016:13:32..2018-05-07%2018:13:03.html))
I guess I can skip opening tickes on the revealed issues, as you are already hunting them down

perost:
@casella: Yes, the regressions are due to forcing evalution of all complex parameter bindings, but the function evaluation is not entirely up to the task yet. I think it's mostly just handling of arrays that's needed.

The report also looks much worse than it actually is since a lot of the models are duplicated many times, and all the regressions are for models that previously failed immediately after the frontend anyway. On the positive side several more MSL models now verify, so even though Hudson says otherwise I think this was good progress.

Also, I get three separate emails linking to the report everytime there are regressions, you don't have to worry about me missing it ;)

casella:
Sure. Just in case they get in your spam folder ;)

PR #2422, opened 2018-05-09
[NF] Fix variability of size() in functions.

perost:
- Limit the variability of size in a function to parameter. The
  variability of the input arguments should be used rather than the
  input parameters, but functions are currently typed without taking the
  input arguments into consideration.
- Propagate origin to Ceval.evalExp, since it might need to type
  expressions when evaluating bindings of constants.
- Removed deprecated NFExpOrigin.mo.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-05-09_15-40-20](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5662/console).

PR #2425, opened 2018-05-09
[NF] Fix variability of size() properly.

perost:
- Partially revert previous commit that propagated origin to
  Ceval.evalExp, and deduce the origin based on where a component is
  declared instead. Package constants should be typed like they're
  class variables, even when used in a function.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-05-09_17-29-03](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5665/console).

PR #2437, opened 2018-05-16
[NF] Record fixes.

perost:
- Generate default record constructors even when they aren't
  explicitly called, since calls might be generated by the backend.
- Remove the 'constructor'.$default suffix that was previously added
  to generated record constructors, since the rest of the compiler
  assumes constructors have the same name as the record itself.
- Use DAE.Function.RECORD_CONSTRUCTOR for record constructors instead
  of DAE.Function.FUNCTION.

### DIFF
--- a/Compiler/NFFrontEnd/NFCall.mo
+++ b/Compiler/NFFrontEnd/NFCall.mo
@@ -194,9 +194,6 @@ uniontype Call
         then
           fail();
     end match;
-
-    // callExp := Expression.CALL(call);
-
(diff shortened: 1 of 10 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-05-16_16-21-53](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5697/console).

PR #2448, opened 2018-05-22
[NF] Improve handling of class modifiers.

perost:
- Replace the whole binding level system of handling nested modifiers
  with keeping a list of a parents for each binding, to be able to
  handle nested modifiers that are a mix of component and classes.
- Changed NFBinding into a package to be able to have constants in it.
- Removed deprecated BindingOrigin module.
- Removed restriction that only types are allowed to have dimensions,
  it works correctly now and the intent of the specification is
  probably that they should be allowed.
- Added better error message when dimension deduction fails due to the
  binding being 'each'.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-05-22_11-42-44](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5711/console).

PR #2457, opened 2018-05-23
[NF] QoL fixes.

perost:
- Fixed order of equations in when-equations.
- Added __OpenModelica_builtinType annotation to allow the $Code types
  in NFModelicaBuiltin to work in both the new and old instantiation,
  so that Interactive doesn't break if NFModelicaBuiltin is loaded at
  start (i.e. by using -d=newInst on the command line instead of
  setting it in the script).


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-05-23_12-12-35](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5721/console).

PR #2469, opened 2018-05-25
[NF] Fix ConnectEquations.makePositiveMaxCall.

perost:
- Fetch the instance node from the flow cref instead of assuming that
  the flow expression is always a cref expressions, which is often
  not the case.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-05-25_14-20-05](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5736/console).

PR #2470, opened 2018-05-25
[NF] Fix call attributes for operator overloading.

perost:
- Create real call attributes for overloaded operator calls, so that
  they get e.g. the correct inline attributes.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-05-25_19-05-05](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5737/console).

PR #2493, opened 2018-06-04
Disable preOptModules removeEqualFunctionCalls

lochel:
### Related tickets

* [ticket:4952](https://trac.openmodelica.org/OpenModelica/ticket/4952)
* [ticket:4881](https://trac.openmodelica.org/OpenModelica/ticket/4881)

### Purpose

The `wrapFunctionCalls` module, which assigns all function calls to cse variables, has a wider scope and in fact makes the `removeEqualFunctionCalls` method obsolete.

### Approach

Remove `removeEqualFunctionCalls` from the list of default modules.

### Open Questions and Pre-Merge TODOs

- [ ] Remove the actual implementation from the source base.


lochel:
@casella I tested it already a few years ago and then some mails failed during index reduction. The problem is that the module `wrapFunctionCalls` is part of the postOptModules and therefore executed after index reduction. So I still expect the same result.

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2018-06-04_14-31-20](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5769/).

PR #2512, opened 2018-06-18
[NF] Function output type fixes.

perost:
- Set the variability of expression dimensions in function outputs to
  constant. This might not be strictly correct, but will cause fewer
  issues than using the actual variability of the output which isn't
  correct either.
- Evaluate function output dimensions too when evaluating functions.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-06-18_11-57-42](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5797/console).

PR #2513, opened 2018-06-18
[NF] Try not evaluating complex parameter bindings.

perost:
- Use the same strategy as the old frontend of moving complex parameter
  bindings that couldn't be split into an initial equation section and
  setting the field to 'fixed = false'.

### DIFF
--- a/Compiler/NFFrontEnd/NFFlatten.mo
+++ b/Compiler/NFFrontEnd/NFFlatten.mo
@@ -300,20 +300,24 @@ protected
   Visibility vis;
   Equation eq;
   list<tuple<String, Binding>> ty_attrs;
+  Variability var;
+  Boolean is_fixed;
 algorithm
   Component.TYPED_COMPONENT(ty = ty, binding = binding, attributes = comp_attr,
     comment = cmt, info = info) := comp;
+  var := comp_attr.variability;
 
   if isSome(outerBinding) then
     SOME(binding) := outerBinding;
+    is_fixed := Binding.isUnbound(binding) and var == Variability.PARAMETER;
   else
     binding := flattenBinding(binding, prefix);
+    is_fixed := false;
   end if;
 
   // If the component is an array component with a binding and at least discrete variability,
   // move the binding into an equation. This avoids having to scalarize the binding.
-  if Type.isArray(ty) and Binding.isBound(binding) and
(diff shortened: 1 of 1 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-06-18_15-46-56](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5798/console).

casella:
Nice try, I guess the devil is in the details 😃 
https://libraries.openmodelica.org/branches/history/newInst/2018-06-18%2015:13:51..2018-06-18%2017:38:14.html

perost:
@casella: Indeed, I basically forgot to use `each` when setting `fixed = true` on array components. And I also realised the commit message says `fixed = false`, but at least the code is correct(ish).

casella:
In fact, I do wonder why on Earth was `each` ever introduced in Modelica, since

1. Everybody forgets about it all the time (including me and you)
2. Dymola doesn't care and it works just fine
3. It would have been easy to introduce automatic fill() operators if the number of dimensions of the right-hand side is higher than the number on the left-hand side.

It seems to me like a completely redundant concept and, as a longtime fan of Ockham's razor, I'd gladly cut it out. Do I miss something?

perost:
>It seems to me like a completely redundant concept and, as a longtime fan of Ockham's razor, I'd gladly cut it out. Do I miss something?
Well, each decides how a modifier is split. So e.g.:
```Modelica
model A
  Real x[2];
end A;

model B
  A a[2](x = {1, 2});
end B;
```
As it's specified right now the modifier on `a` would be split so that you'd end up with:
```Modelica
a[1].x = 1;
a[2].x = 2;
```
Adding `fill()` operations would finally give the flat model:
```Modelica
a[1].x[1] = 1;
a[1].x[2] = 1;
a[2].x[1] = 2;
a[2].x[2] = 2;
```
In contrast, if the modifier would have had an `each` the flat model would instead be:
```Modelica
a[1].x[1] = 1;
a[1].x[2] = 2;
a[2].x[1] = 1;
a[2].x[2] = 2;
```
It might be possible to specify modifiers without using `each` somehow, but as it is I think it would just make it even more complicated to implement if `each` was just removed. Also, this particular model doesn't work in Dymola, so Dymola does care about `each` at least sometimes.

PR #2515, opened 2018-06-19
[NF] Operator overloading fixes.

perost:
- Handle +/- with arrays when no exact matching operator is defined.
- Fix the matching so that the operator overloading can use operator
  functions that take operator record arrays as argument.
- Move some code from TypeCheck to OperatorOverloading.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-06-19_16-31-47](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5800/console).

PR #2523, opened 2018-06-25
[NF] Update NFModelicaBuiltin.

perost:
- Use the same inline definition of 'rem' as in the old frontend,
  since inlining now works and the backend doesn't handle 'rem'.
- Port recent changes of ModelicaBuiltin to NFModelicaBuiltin.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-06-25_13-04-25](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5810/console).

PR #2532, opened 2018-06-27
[NF] Tuple and array equation improvements.

perost:
- Simplify `cat` calls.
- Simplify "tuple subscripted" arrays by making each array element a
  tuple subscript.
- Move the check whether an array equation should be scalarized or not
  from the typing to the scalarization, so it's done after
  simplification (which might affect the decision).
- Evaluate "tuple subscripted" expressions in Ceval.

### DIFF
--- a/Compiler/NFFrontEnd/NFCeval.mo
+++ b/Compiler/NFFrontEnd/NFCeval.mo
@@ -160,6 +160,12 @@ algorithm
         if EvalTarget.isRange(target) then
           Expression.RANGE(exp.ty, exp1, oexp, exp3) else evalRange(exp1, oexp, exp3);
 
+    case Expression.TUPLE()
+      algorithm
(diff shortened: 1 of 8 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-06-27_15-02-33](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5819/console).

casella:
I guess this was the cause of quite some regressions, see https://libraries.openmodelica.org/branches/history/newInst/2018-06-26%2021:02:57..2018-06-27%2016:08:39.html

casella:
It seems that https://github.com/OpenModelica/OMCompiler/commit/561548584 more than offsets this 👍

PR #2553, opened 2018-07-03
Fix the FMU resource path on Win32

sjoelund:
Do not strip all characters if there is a `:` in the path; only strip
leading `/`.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-07-03_08-49-58](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5851/console).

PR #2554, opened 2018-07-03
[NF] Even better extends lookup fix.

perost:
- Don't return the first scope if it has the same name as what's being
  looked for, in that case the name needs to be looked up in the
  enclosing scope and instantiated as a package.


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2018-07-03_14-27-38](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5854/).

PR #2555, opened 2018-07-03
[NF] Revert to first base class lookup fix.

perost:
- Just checking that the scope is a class is probably the best way
  after all.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-07-03_14-56-54](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5855/console).

PR #2556, opened 2018-07-03
[NF] Fix simplification of sum/product.

perost:
- Use the array element type of the argument when creating default
  expressions, since the result should be a scalar.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-07-03_15-11-56](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5856/console).

PR #2568, opened 2018-07-06
Fix FMU build on Windows and other FMI fixes

adrpo:
- use fmuTargetName instead of fileNamePrefix to remove the fmu
- use rm -f if System.removeDirectory doesn't work on Windows
- compile with static libstdc++ and pthreads
- don't add extra dlls to the binary directory
- fix resource copy on Windows


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-07-06_03-41-35](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5872/console).

PR #2588, opened 2018-08-03
Optionally cache stackBottom in mmc_init_stackoverflow()

atrosinenko:
... in case calling thread does not plan to use coroutines, etc.

This can make OMEdit start 3 times faster on Linux. 

Required by OpenModelica/OMEdit#173

### DIFF
--- a/SimulationRuntime/c/meta/meta_modelica_segv.c
+++ b/SimulationRuntime/c/meta/meta_modelica_segv.c
@@ -231,9 +231,22 @@ void init_metamodelica_segv_handler()
   sigfillset(&segvset);
 }
 
+
+#ifdef HAS_STACK_CACHING
+__thread int mmc_cache_stack = 0;
+#endif
 void mmc_init_stackoverflow(threadData_t *threadData)
 {
-  threadData->stackBottom = getStackBase();
+#ifdef HAS_STACK_CACHING
+  static __thread void *stackBottom;
+  if (!stackBottom || !mmc_cache_stack)
+#else
+  void *stackBottom;
+#endif
+  {
+    stackBottom = getStackBase();
+  }
+  threadData->stackBottom = stackBottom;
 }
 
 #else
--- a/SimulationRuntime/c/meta/meta_modelica_segv.h
+++ b/SimulationRuntime/c/meta/meta_modelica_segv.h
@@ -57,6 +57,11 @@ static inline void mmc_init_stackoverflow(threadData_t *threadData)
 void mmc_init_stackoverflow(threadData_t *threadData);
 #endif
 
+#ifdef __linux
+#define HAS_STACK_CACHING
+extern __thread int mmc_cache_stack;
+#endif
+
 #ifndef __has_builtin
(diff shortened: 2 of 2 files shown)


atrosinenko:
Is it a viable solution?

adrpo:
We haven't checked yet. I will at least run the testsuite on it and see how it goes.

atrosinenko:
I fear, if this PR would introduce some issues, they can be caught only in OMEdit, since this behavior is off by default (at least I expected that).

adrpo:
I'm not sure about portability, we need to run this code on several platforms with different architectures and compilers. I guess we could allow it on Linux and see if people complain. Does it work fine in your tests?

atrosinenko:
Not sure it works absolutely fine, but it starts fast :) I used this primarily for developing OMEdit on a kind of slow laptop to not wait for starting OMEdit after recompile too much. I will try to look at it more thoroughly on weekend. This most probably can be delayed till next release cycle if needed. Please excuse me for disturbance.

atrosinenko:
Tested this with the following program (copied function from OM run-time):

```c
#define _GNU_SOURCE
#include <pthread.h>
#include <stdio.h>
#include <assert.h>

/* From OpenModelica run-time */
static void* getStackBase() {
  /* Warning: These functions are highly non-portable and are recommended to not be used.
   * We only tested them on Linux and OSX.
   * On OSX we get the top of the stack and the size
   * On Linux we get the bottom, so we don't need the size... YMMV
   */
  void* stackBottom;
  pthread_t self = pthread_self();
#if !defined(__APPLE_CC__)
  size_t size = 0;
  pthread_attr_t sattr;
  pthread_attr_init(&sattr);
  pthread_getattr_np(self, &sattr);
  assert(0==pthread_attr_getstack(&sattr, &stackBottom, &size));
  assert(stackBottom);
  pthread_attr_destroy(&sattr);
#else
  void* addr = pthread_get_stackaddr_np(self);
  size_t size = pthread_get_stacksize_np(self);
  stackBottom = (void*) (((long)addr) - size);
  if (pthread_main_np()) {
    char str[256];
    size_t size = sizeof(str);
    int ret = sysctlbyname("kern.osrelease", str, &size, NULL, 0);
    if (0==ret && str[0]=='1' && str[1]=='3' && str[2]=='.') {
      /* OSX Mavericks returns a wrong stack size... Assume default 8MB */
      stackBottom = (void*) (((long)addr) - 8 * 1024 * 1024);
    }
  }
#endif
  assert(size > 128*1024);
  return stackBottom + 64*1024;
}

void func() {
  char x[128];
  void *p = getStackBase();
  fprintf(stderr, "%p\n", p);
  func();
}

int main() {
  func();
}
```

On my system it shown me 47540 lines _with identical pointer values_ before crash which seems good with respect to this PR.

Another, more robust, option for this may be to pass old `threadData` to `mmc_init_stackoverflow` and copy the stack bottom from there (if present) -- this would probably handle coroutines transparently but would change API/ABI.

atrosinenko:
Reworked as #2884

PR #2598, opened 2018-08-10
Allow the compliance suite tests to crash

sjoelund:
If a test always crashes or fails, we can still test for this behaviour
even if it is a really bad result.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-08-10_07-53-43](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5929/console).

PR #2603, opened 2018-08-13
[NF] Fix instantiation of derived classes.

perost:
- Fix Typing.typeClass so that it correctly handles the case where the
  root class is a short class definition.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-08-13_10-00-11](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5938/console).

PR #2616, opened 2018-08-23
expand type scope fully if record (ticket:5075)

adrpo:


adrpo:
This sort of works if we fully expand the path for records. However it seems that while in the DAE the types seem to be correct, when generating equations other record names seem to be used which are not available. See more in the ticket.

adrpo:
Closing this as @perost fixed all this stuff already.

PR #2625, opened 2018-09-01
Fix String inputs/outputs handling for FMUs

atrosinenko:
Fix compilation of FMUs with String inputs/outputs. Frankly speaking, I'm not experienced at all with FMU runtime, but analyzing compiler errors and generated code I have made this 4-line modification. It seems that now it successfully compiles
```modelica
model Test
  output Real x;
  output Real y;
  input Real w;
  input String inStr;
  input Boolean b;
  output String outStr;
  output String msg;
initial equation
  x = 1;
equation
  der(x) = y;
  der(y) = -w * w * x;
  outStr = inStr;
  msg = (if x < 0 then "negative" else "positive") + " & " + (if b then "true" else "false");
end Test;
```
... with command like `buildModelFMU(Test, "1.0", "me");` and runs under FMUChecker with input
```csv
"time";"w";"b"
0.0;3.1415926;0
0.7;3.1415926;1
```
... in 1.0 ME and 2.0 CS modes. With 2.0 ME, it seems that something is broken with setting Booleans (removing the `"b"` column helps) but this error presents before this PR too.

Cannot test _setting_ String inputs right now, since FMUChecker does not support this.

PS: Hope compliance tests will catch regressions if any... :)

### DIFF
--- a/Compiler/Template/CodegenC.tpl
+++ b/Compiler/Template/CodegenC.tpl
@@ -1654,7 +1654,7 @@ template functionInput(SimCode simCode, ModelInfo modelInfo, String modelNamePre
     {
       TRACE_PUSH
 
-      <%vars.inputVars |> SIMVAR(__) hasindex i0 =>
+      <%vars.inputVars |> SIMVAR(name=name, type_=T_REAL()) hasindex i0 =>
         '<%cref(name)%> = data->simulationInfo->inputVars[<%i0%>];'
         ;separator="\n"
       %>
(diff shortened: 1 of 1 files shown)


atrosinenko:
Setting String inputs seems to be OK in my tests. This PR should probably be quite useful and easily auto-testable for regressions (assuming testsuite checks FMUs), though may require enabling building some new models to see if new functionality works well, too.

adrpo:
I will run the testsuite on this. Basically what you do is ignore all other inputs than Real. While this works for now, I guess we should try to properly fix this by adding code to set String and Boolean as well. I will open a ticket on this so we don't forget about it.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-09-10_19-20-25](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6000/console).

adrpo:
See also: https://trac.openmodelica.org/OpenModelica/ticket/5113

PR #2628, opened 2018-09-03
[NF] Improve subscripting of expressions.

perost:
- Improved Expression.applySubscripts so that multiple subscripts as
  well as slice and : subscripts are handled correctly.
- Changed Expression.SUBSCRIPTED_EXP to use Subscript for subscripts
  instead of Expression, to better handle slicing.

### DIFF
--- a/Compiler/NFFrontEnd/NFCall.mo
+++ b/Compiler/NFFrontEnd/NFCall.mo
@@ -64,6 +64,7 @@ import Prefixes = NFPrefixes;
 import TypeCheck = NFTypeCheck;
 import Typing = NFTyping;
 import Util;
+import Subscript = NFSubscript;
 
 public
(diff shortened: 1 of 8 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-09-03_16-28-58](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5979/console).

casella:
Looking at the [regression report](https://libraries.openmodelica.org/branches/history/newInst/2018-09-03%2003:07:39..2018-09-03%2021:07:39.html), there are over 300 regressions and not a single improvement related to this commit.

Was it commit trying to fix any issue reported on the testsuite? If so, why doesn't it work?

PR #2634, opened 2018-09-06
[NF] SimplifyExp improvements.

perost:
- Added trivial simplifications of binary expressions, such as 0*e=0.
- Added simplification of transpose of an array expression.
- Turned off expansion of expressions containing calls to functions with
  array return type when -d=nfExpandOperations is used (on by default),
  to better mimic the old frontend and avoid unnecessary calls.
- Rearranged SimplifyExp and ExpandExp a bit so that expressions are
  always simplified before being expanded with -d=nfExpandOperations.
- Try to expand the argument in Expression.promoteExp to avoid getting
  promote calls in the flat model that the backend can't handle.

### DIFF
--- a/Compiler/NFFrontEnd/NFExpandExp.mo
+++ b/Compiler/NFFrontEnd/NFExpandExp.mo
@@ -465,7 +465,7 @@ public
       end if;
 
       if expanded then
-        outExp := expandBinaryElementWise2(exp1, op, exp2, makeBinaryOp);
+        outExp := expandBinaryElementWise2(exp1, op, exp2, SimplifyExp.simplifyBinaryOp);
       else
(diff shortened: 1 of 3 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-09-06_13-50-22](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/5992/console).

casella:
> Added trivial simplifications of binary expressions, such as 0*e=0
Gosh, I thought this had already been in place, as well as 1*e = e, 0 + e = e, e/1 = 0 and the like

PR #2640, opened 2018-09-10
use LONG_MAX for integer in FMU min/max

adrpo:


adrpo:
We might need to include <limits.h>, let's see if is not already included.

PR #2647, opened 2018-09-14
[NF] Fix typing of min/max.

perost:
- Do manual typing of min/max instead of relying on polymorphic types
  that doesn't work correctly.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-09-14_10-56-30](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6012/console).

PR #2659, opened 2018-09-17
[NF] Fix simplification of transpose.

perost:
- Check that the second dimension of the argument to transpose is also
  expanded before trying to simplify the call.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-09-17_11-20-40](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6031/console).

PR #2661, opened 2018-09-17
[NF] Variability fixes.

perost:
- Change the variability of size to be constant, except for size of
  : in functions where it's discrete.
- Disable variability checks for bindings in functions. It's not well
  specified how it should work, and the variability doesn't really need
  to match when initializing local variables in a function anyway.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-09-17_13-13-07](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6033/console).

PR #2676, opened 2018-09-24
[BE] added debug flag to enable error asserts from  min/max attributes

wibraun:
- debug flag warnMinMax is default to true, because the desired
   behavior is an error assert
 - also check min/max assert in the initialization
 - fixing ticket:4985


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-09-24_15-58-08](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6059/console).

PR #2681, opened 2018-09-28
Upgrades icon generation script to Python 3

dietmarw:


dietmarw:
@hkiel Could you have a quick look?

sjoelund:
Python 2/3 differences?

Examples/generate_icons.py
```diff
@@ -254,7 +254,7 @@ def getGraphicsForClass(modelicaClass):
             graphicsObj['fillPattern'] = g[11]
             graphicsObj['lineThickness'] = float(g[12])
             graphicsObj['extent'] = [[float(g[13]), float(g[14])], [float(g[15]), float(g[16])]]
-            graphicsObj['textString'] = g[17].strip('"').encode('utf-8')
```
hkiel: Seems to work fine without encoding, the `decode()` there was wrong.
Why not remova also `encode()` in line 300?

hkiel:
Without encode() output looks fine with 2.7. I checked `Modelica.ComplexBlocks.ComplexMath.Bode`

hkiel:
The symlink in line 1137 could be protected by
```
if not os.path.exists(filename):
    os.symlink(hashName, filename)
else:
    logger.warning('Target file {0} already exists'.format(filename))
```
for dumb (= case insensitive) filesystems

Examples/generate_icons.py
dietmarw: I overlooked that one. But there are two more and I'm not 100% sure of the impact. So perhaps somebody else can have a look at those.

dietmarw:
> The symlink in line 1137 could be protected by
> 
> ```
> if not os.path.exists(filename):
>     os.symlink(hashName, filename)
> else:
>     logger.warning('Target file {0} already exists'.format(filename))
> ```
> for dumb (= case insensitive) filesystems

That would just warn but not solve the issue right?

dietmarw:
I found another issue. A text string is not correctly scaled. See https://build.openmodelica.org/Documentation/Icons/Complex.svg
which really should look like:
image

dietmarw:
You should be able to push to my PR.

hkiel:
> > for dumb (= case insensitive) filesystems
> 
> That would just warn but not solve the issue right?

 Right. But you cannot do anything about it, and IMO it should not fail. At least you get a warning.

hkiel:
> You should be able to push to my PR.

Cannot push: `remote: Permission to dietmarw/OMCompiler.git denied to hkiel`

dietmarw:
OK in that case push this to your branch and I can cherry pick it from there.

hkiel:
I created #2684 with the fixes for text height.

We should think about moving from deprected `optparse` to `argparse`

dietmarw:
@sjoelund Could you merge this PR and #2684 after that?

dietmarw:
@hkiel 	I'm always for updating deprecated code.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-10-01_16-15-15](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6070/console).

PR #2683, opened 2018-10-01
increase the timeout 600s is not enough for MSL icons

adrpo:


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-10-01_03-37-33](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6066/console).

hkiel:
Maybe also the generated files should be created in a directory tree, as the number of files per directory is really large and I suspect a certain slowdown by this (script seems to slow down with time).

PR #2689, opened 2018-10-03
KINLapackCompletePivotingFree no return in some sundials versions

Murmele:
KINLapackCompletePivotingFree has no return value if the major version is not 2 and the minor version not > 6


sjoelund:
We do compile against the Sundials version provided within OpenModelica though. If it for some reason does not, that should be fixed instead of trying to cover up the problem

Murmele:
At the moment I try to create a flatpak package for openmodelica. But in the git repository there is no sundials library

sjoelund:
https://github.com/OpenModelica/OMCompiler-3rdParty/tree/master/sundials has it

sjoelund:
You shouldn't need to. Our docker image of build dependencies (docker.openmodelica.org/build-deps:v1.13) does not include sundials and it compiles the optional C++ run-time just fine.

PR #2690, opened 2018-10-03
[NF] Move general evaluation to separate phase.

perost:
- Added new module NFEvalConstants which evaluates constants and
  structural parameters after the flattening, instead of doing it during
  the typing which causes lots of issues.
- Moved the construction of the function tree to after the evaluation
  and simplification phases, to avoid collecting functions that have
  been optimized out.
- Fixed Ceval.evalCref so that subscripts are applied correctly.
- Fixed EvalFunction so that it fails if an assertion is triggered,
  instead of returning the current value of the outputs.
- Improved subscripting of if-expressions by subscripting the branches
  instead of the whole expression.
- Added check that terminate is not used in a function context,
  and removed the handling of terminate in EvalFunction.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-10-03_15-26-54](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6077/console).

PR #2691, opened 2018-10-03
[NF] don't try to evaluate external functions in the backend

adrpo:
In NFModelicaBuiltin.mo the builtin function max has no elements and because of that it was considered a record in the backend! Now we skip external functions and check if is actually a record.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-10-04_01-59-17](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6078/console).

PR #2692, opened 2018-10-04
Activate for-equation in backend and Cpp code generation

rfranke:
Basic assumption: one for-equation defines one array variable. This makes sense along with
```
omc -d=newInst,-nfScalarize
```
Many backend modules still fail. But simple models do simulate and can be exported as FMU with arrays and for loops.

### DIFF
--- a/Compiler/BackEnd/BackendDAECreate.mo
+++ b/Compiler/BackEnd/BackendDAECreate.mo
@@ -424,6 +424,13 @@ algorithm
         then
           ();
 
+      // for equation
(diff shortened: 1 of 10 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-10-04_12-35-38](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6081/console).

PR #2701, opened 2018-10-08
[NF] Fix handling of Evaluate annotation.

perost:
- Only consider the Evaluate annotation for parameters as per the
  specification, otherwise non-parameters might be turned into parameters.

### DIFF
--- a/Compiler/NFFrontEnd/NFInst.mo
+++ b/Compiler/NFFrontEnd/NFInst.mo
@@ -2931,7 +2931,7 @@ algorithm
       algorithm
         // @adrpo: if Evaluate=true make the parameter a structural parameter
         // only make it a structural parameter if is not constant, duh!, 1071 regressions :)
-        if Component.getEvaluateAnnotation(c) and (Component.variability(c) > Variability.STRUCTURAL_PARAMETER) then
+        if c.attributes.variability == Variability.PARAMETER and Component.getEvaluateAnnotation(c) then
           InstNode.updateComponent(Component.setVariability(Variability.STRUCTURAL_PARAMETER, c), node);
         end if;
 


OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2018-10-08_10-44-32](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6091/).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-10-08_11-02-32](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6092/console).

casella:
This is of course reasonable, but why should people put the `Evaluate = true` annotation on anything else then a parameter? Did this actually happen in some covered library?

perost:
> This is of course reasonable, but why should people put the `Evaluate = true` annotation on anything else then a parameter? Did this actually happen in some covered library?

Yes, this was causing many Buildings-related models to be unbalanced due to continuous variables being turned into parameters (since we handle `Evaluate = true` by making the component a structural parameter). See e.g. BuildSysPro.IBPSA.Controls.Continuous.LimPID for a model where Evaluate is used on several non-parameters (`Dzero`, `Izero`, `y_reset_internal`). The specification explicitly says that Evaluate only has an effect for components with the prefix parameter, so now we just skip the check if a component isn't a parameter.

casella:
> > This is of course reasonable, but why should people put the `Evaluate = true` annotation on anything else then a parameter? Did this actually happen in some covered library?
> 
> Yes, this was causing many Buildings-related models to be unbalanced due to continuous variables being turned into parameters (since we handle `Evaluate = true` by making the component a structural parameter). See e.g. BuildSysPro.IBPSA.Controls.Continuous.LimPID for a model where Evaluate is used on several non-parameters (`Dzero`, `Izero`, `y_reset_internal`). 

Aha. Classical example of Murphy's law at work (the real one, not the one of the jokes)

PR #2704, opened 2018-10-08
generate hashed SVGs in subdirs

hkiel:
put hash-named SVGs into 16 subdirs according to first letter to reduce number of files in one directory

### DIFF
--- a/Examples/generate_icons.py
+++ b/Examples/generate_icons.py
@@ -53,12 +53,6 @@ def classToFileName(cl):
   """
   return cl.replace("/","Division").replace("*","Multiplication")
 
-# Bitmap
-#   extends GraphicItem
-#   extent
-#   fileName
-#   imageSource
-
 exp_float = '[+-]?\d+(?:.\d+)?(?:e[+-]?\d+)?'
 
 element_id = 0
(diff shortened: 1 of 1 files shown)


hkiel:
Would it make sense to create hierarchy of files to further reduce the number?
`Modelica.Mechanics.Rotational.Examples.Utilities.SpringDamper.svg` -> `Modelica/Mechanics/Rotational/Examples/Utilities/SpringDamper.svg`

adrpo:
It would maybe make it easier to find out what class the svg belongs to, now we have generated hashes that do not tell anything.  But you would need to start with the library version number first to make things unique:  `Modelica_latest/Modelica/Mechanics/Rotational/Examples/Utilities/SpringDamper.svg`

hkiel:
The hashes are there to save space, so that e.g. the Example icon (green triangle) can be re-used for all example models.
Anyways, with the current scripting it seems that you can only generate documentation of one version of the library. This needs to be changed first. Then we could modify the script to only generate the documentation if the library changed (or the scripts of course).

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-10-10_10-21-04](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6101/console).

PR #2767, opened 2018-11-06
[NF] Remove bad NFCeval optimization.

perost:
- Removed the caching of record field bindings that are created from the
  record instance's binding. It's not safe and causes subscripts to be
  dropped in some cases, and the record instance's binding is cached now
  anyway.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-11-06_12-41-51](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6181/console).

PR #2771, opened 2018-11-07
[NF] Evaluate parameter subscripts.

perost:
- Mark parameter subscripts as structural so that they're evaluated.
  This should ideally not be necessary, but the backend relies on it
  since the old frontend does it.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-11-07_14-04-01](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6190/console).

PR #2784, opened 2018-11-12
Fix inline type checking.

perost:
- Allow record types to have different names when checking that the type
  of the call to be inlined and its output is the same.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-11-12_13-00-15](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6206/console).

PR #2796, opened 2018-11-23
Fix anyString for SOME().

perost:
c54cb4b24af5feed52e444e5923d47c93e382341 implemented support for handling cycles in anyString, but it changed the printing of SOME() such that it tries to print the seemingly uninitialized variable data. This commit reverts that change, which doesn't seem to affect the cycles detection but fixes the issue that SOME() is not printed correctly.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-11-23_16-41-39](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6225/console).

PR #2797, opened 2018-11-23
[NF] Initial partial function application support.

perost:
- Implemented handling of partial function applications.
- Changed TypeMatch.matchExpressions to handle type checking/casting by
  itself instead of relying on matchTypes, since matchTypes assumes a
  strict actual/expected relation between the types which is not the
  case for matchExpressions. This could e.g. lead to inconsistent
  boxing/unboxing of expressions depending on which side of a binary
  expression a boxed expression appeared on.
- Replaced all usage of intBitOr/intBitAnd for ExpOrigin flags with
  ExpOrigin.setFlag/flagSet.

### DIFF
--- a/Compiler/NFFrontEnd/NFBuiltinCall.mo
+++ b/Compiler/NFFrontEnd/NFBuiltinCall.mo
@@ -531,7 +531,7 @@ protected
     end if;
 
     // pre/change may not be used in a function context.
-    if intBitAnd(origin, ExpOrigin.FUNCTION) > 0 then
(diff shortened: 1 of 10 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-11-23_17-56-00](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6226/console).

PR #2803, opened 2018-11-28
[NF] Clone inherited classes.

perost:
- Clone inherited classes so that they get unique caches, to prevent e.g.
  the same record constructor instance from being shared between classes
  where the record has different modifiers or use package constants with
  different bindings.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-11-28_15-22-35](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6232/console).

PR #2807, opened 2018-12-02
[Cpp] Implement flags for input and output path, ticket:4773

rfranke:
This follows up 1be2a07fd09f6bc5eab46ffcd568011d99d007be that implemented
these flags to the C runtime.

Both flags are added to GlobalSettings and default to the path of the
compiled model. The flag inputPath replaces the previously hard coded
path in OMCpp<Model>Initialize.cpp.

The flag outputPath is not yet used (the results file has an own flag).
These flags make sense once more that one input or output file will be used.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-12-02_17-37-04](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6248/console).

PR #2813, opened 2018-12-03
[Cpp] Don't use outputPath for results file, ticket:4773

rfranke:
The results file has its own call argument -F (or -r for the C runtime)
that is typically used with an absolute file name, including path.
A results file with relative name should be placed in the current
directory according to discussions at dev meeting on 12/03, 2018.

Moreover raise an error if open of the results file fails.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-12-03_20-25-28](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6255/console).

PR #2815, opened 2018-12-06
[NF] Improve functions called through components.

perost:
- Clone all classes except operators (because operators are special
  and cloning them causes typing loops) when instantiating a class
  tree, to ensure that classes inside of instantiated components have
  unique caches.
- Remove cloning of extends nodes, since cloning all classes makes
  that unnecessary.
- Implemented Statement.toString for debugging.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-12-06_15-16-35](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6259/console).

PR #2819, opened 2018-12-07
[NF] Various fixes.

perost:
- Remove the type in Call.CallAttributes, the call itself already
  contains the same type.
- Improve subscripting of calls such as pre(), where the argument can
  be subscripted instead of the whole call expression.
- Fix Type.copyDims so that it doesn't create an array with no
  dimensions if the source type is a scalar type.
- Add simplification 'if ... then x else x' => 'x'.

### DIFF
--- a/Compiler/NFFrontEnd/NFCall.mo
+++ b/Compiler/NFFrontEnd/NFCall.mo
@@ -70,7 +70,6 @@ import Operator = NFOperator;
 public
   uniontype CallAttributes
     record CALL_ATTR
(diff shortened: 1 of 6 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-12-07_16-16-34](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6263/console).

PR #2822, opened 2018-12-10
[NF] Relax restriction on input/output prefixes.

perost:
- Allow an input/output prefix on a component with a short class type
  that also has an input/output prefix, as long as they are the same.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-12-10_13-54-45](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6267/console).

PR #2826, opened 2018-12-12
[NF] Record constructor fixes.

perost:
- Only add constructor inputs to the list of field names when
  constructing ComplexType.RECORD types.
- Make non-modifiable fields protected when creating the DAE type, like
  the old frontend does.
- Propagate visibility to Component.Attributes.toDAE so we get the
  actual visibility in the DAE.ATTR.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2018-12-12_15-54-59](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6271/console).

PR #2834, opened 2018-12-13
Fix typos

lochel:


casella:
@lochel please apply this to maintenance/v.13, otherwise it won't make it into the final release

PR #2846, opened 2019-01-03
Implement *_get_5D(...) array accessors

atrosinenko:
This PR should technically fix the possibility to at least build models like `Buildings.HeatTransfer.Windows.BaseClasses.Examples.AbsorbedRadiation` and looks similar to what already exists.

I have rephrased the `getIndex_nD` function implementation, so they should be easier to read and, possibly, faster to compute (but this highly depends on what the optimizer already does). This may interfere with some corner cases like integer overflow, but these are probably UB-like cases anyway. In case the old implementation was written in such a way for a purpose, then maybe it would be worth to write some comment like

```c
// Index = i0 * dim[1] * dim[2] * ... * dim[n] + i1 * dim[2] * ... * dim[n] + ... + i{n-1} * dim[n] + i{n}
```

### DIFF
--- a/SimulationRuntime/c/util/base_array.h
+++ b/SimulationRuntime/c/util/base_array.h
@@ -37,8 +37,9 @@
 #include "omc_msvc.h"
 
 static OMC_INLINE size_t getIndex_2D(_index_t *dim, int i, int j) {return i*dim[1]+j;}
-static OMC_INLINE size_t getIndex_3D(_index_t *dim, int i, int j, int k) {return i*dim[1]*dim[2]+j*dim[2]+k;}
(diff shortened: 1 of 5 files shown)


atrosinenko:
Hmm... There are references to `string_get_2D` in `KeyWordIO.Examples.ReadStringCSV` as well...

atrosinenko:
Adapted `real_get_nD` for strings, hope this copy-paste is not agains CLA. :)

hkiel:
Your patch looks good. I wonder if 6D or 7D is also needed in some other model...
I started hudson job to run the testsuite on this and merge if successful.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2019-01-03_17-52-50](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6299/console).

PR #2848, opened 2019-01-07
[NF] Fix DAE conversion of / operator.

perost:
- Element-wise division of arrays should use DAE.DIV_ARR and not
  DAE.DIV.


hkiel:
Should this also be applied to POW/POW_ARR ?

perost:
> Should this also be applied to POW/POW_ARR ?

Good point, but POW_ARR is used for the `matrix ^ scalar` case instead. But now I see that there's actually a POW_ARR2 that the old frontend uses for element-wise `^`, so I'll fix the code to generate that instead.

PR #2852, opened 2019-01-09
[NF] Various fixes.

perost:
- Don't cache bindings created in NFCeval.makeComponentBinding if the cref
  has subscripts, it might not be safe.
- Don't ignore subscripts in connections when building the connection
  graph.
- Don't expand record in Expression.extendArrExp. Doing so causes record
  crefs to be expanded into arrays of record fields, which causes
  BackendVarTransform to make invalid replacements.

### DIFF
--- a/Compiler/FrontEnd/Expression.mo
+++ b/Compiler/FrontEnd/Expression.mo
@@ -12051,6 +12051,7 @@ public function expandExpression
  "
 
   input DAE.Exp inExp;
+  input Boolean expandRecord;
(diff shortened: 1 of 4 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2019-01-09_12-15-57](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6316/console).

PR #2861, opened 2019-01-10
[NF] Ceval/SimplifyExp improvements.

perost:
- When building record bindings, only evaluate structural parameter
  fields and not the whole generated expression (which might contain
  non-constant expressions).
- Added simplification of and/or expressions where only one side is
  true/false (for example 'true or e => true').


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2019-01-10_15-23-46](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6332/console).

PR #2864, opened 2019-01-11
Handle record better in lowerArrayEqn.

perost:
- Move the case that handles arrays of records to the top in
  BackendDAECreate.lowerArrayEqn, to avoid array equations involving
  records being turned into arrays of normal non-complex equations.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2019-01-11_15-18-35](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6336/console).

PR #2884, opened 2019-01-19
Use the stack bottom value from the old threadData, if available

atrosinenko:
On linux, fetching the stack bottom involves huge amounts of user and kernel CPU time. Assuming the  stack bottom does not change over time, copy it from the old threadData structure.

This PR uses the same assumptions as #2588 but should work with coroutines as well, so it is enabled on Linux automatically and does not requires changes in programs that use OMCompiler.

It speed ups starting of OMEdit by ~2.5x on my computer and fetching the contents of OpenModelica/Scripting by about 15%.

Replaces #2588 and OpenModelica/OMEdit#173.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2019-01-19_13-31-12](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6364/console).

adrpo:
Any idea if we can apply the same on MacOS?

adrpo:
@atrosinenko: can you close the other PRs that this replaces if not already done so?

sjoelund:
MacOS already has working pthread_getspecific. It wouldn't work on any platform where pthread_getspecific is stubbed away by global variables. It shouldn't affect regular OMC because MMC_TRY_TOP is called pretty much... once?

PR #2888, opened 2019-01-22
add fmiinitial dependency list in fmiexport

arun3688:
### Purpose
This implements FMI export for initial unknowns in ModelStructure  of the ModelDecription.xml

### Files Changed 
CodegenFMUCommon.tpl : 
Check for Variable "ísfixed" inorder to set the initial to be calculated or exact 

SimCodeUTil.mo :
Added Function getFMIInitialDep which takes the initialization DAE and filter the variables based on causality = "output " and fixed = "false" to be added to the initialunknowns in modelstructure


lochel:
@arun3688 Did you test `testsuite\omsimulator\DualMassOscillator_me.mos` without using `--ignoreInitialUnknowns=true`?

PR #2905, opened 2019-01-28
[NF] Partially revert 8fcf925.

perost:
- The equations shouldn't be reverted when replacing an if-equation with
  its first branch.
- Disable simplification of for-loops over a one-sized range, it causes
  some models to become unbalanced for unknown reasons.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2019-01-28_18-59-42](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6394/console).

PR #2909, opened 2019-01-30
[NF] Make Real range generation more robust.

perost:
- Generate Real ranges using start+step*i, instead of relying on the
  code generation that uses repeated addition which causes unnecessary
  inaccuracies.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2019-01-30_13-26-51](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6400/console).

PR #2917, opened 2019-02-04
Fix code generation for function outputs.

perost:
- Fix code generation for array output function parameters with literal
  default values. Such outputs were previously set to be equal to the
  shared literal, which caused a segfault if the array was then assigned
  to. This fix makes sure that a new array is created from the literal
  instead.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2019-02-04_17-22-20](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6415/console).

PR #2927, opened 2019-02-11
[NF] Fix type when casting arrays.

perost:
- When casting arrays in NFSimplifyExp.simplifyCast, don't just cast
  the elements but update the type of the array too.

### DIFF
--- a/Compiler/NFFrontEnd/NFSimplifyExp.mo
+++ b/Compiler/NFFrontEnd/NFSimplifyExp.mo
@@ -667,11 +667,19 @@ function simplifyCast
   output Expression castExp;
 algorithm
   castExp := match (ty, exp)
+    local
+      Type ety;
+
     case (Type.REAL(), Expression.INTEGER())
       then Expression.REAL(intReal(exp.value));
 
     case (Type.ARRAY(elementType = Type.REAL()), Expression.ARRAY())
-      then Expression.mapArrayElements(exp, function simplifyCast(ty = Type.REAL()));
+      algorithm
+        ety := Type.unliftArray(ty);
+        exp.elements := list(simplifyCast(e, ety) for e in exp.elements);
+        exp.ty := Type.setArrayElementType(exp.ty, Type.arrayElementType(ty));
+      then
+        exp;
 
     else Expression.CAST(ty, exp);
   end match;


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2019-02-11_18-06-50](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6433/console).

casella:
Wow, this commit had a very positive impact on the [testsuite](https://libraries.openmodelica.org/branches/history/newInst/2019-02-11%2011:39:05..2019-02-11%2021:05:14.html), particularly regarding the cpp runtime. @rfranke, @niklwors, I guess this is good news for you.

It also fixed some issues with MultiBody models whose root causes I was trying hard to understand, RevoluteConstraint, SphericalConstraint, UniversalConstraint, RollingWheel, and thus also ticket [#5328](https://trac.openmodelica.org/OpenModelica/ticket/5328)

perost:
> Wow, this commit had a very positive impact on the [testsuite](https://libraries.openmodelica.org/branches/history/newInst/2019-02-11%2011:39:05..2019-02-11%2021:05:14.html), particularly regarding the cpp runtime. @rfranke, @niklwors, I guess this is good news for you.
> 
> It also fixed some issues with MultiBody models whose root causes I was trying hard to understand, RevoluteConstraint, SphericalConstraint, UniversalConstraint, and RollingWheel.

Yes, fixing those MultiBody models was the intention, improving the cpp runtime coverage was just a nice bonus. 

I noticed those MultiBody models where triggering the warning message about iteration variables with default zero start attribute (the `-d=initialization` one), which wasn't the case with the old frontend. I did some digging and figured out that the inlining failed on one of the functions, because an Integer array in the function had been typecast to Real without changing the type of the array itself. And the inlining checks that the type of the expression is the same before and after inlining, and since that wasn't true in this case it aborted. So in the end one of the equations wasn't inlined quite as much as with the old frontend, and ended up breaking the simulation for some reason.

casella:
I may have anther similar case for you, see [#5332](https://trac.openmodelica.org/OpenModelica/ticket/5332). In this case the issue is not inlining but constant evaluation.

PR #2941, opened 2019-02-25
[NF] Implementation of expandable connectors.

perost:
- Added new module ExpandableConnectors that handles expandable
  connections.
- Changed Prefixes.ConnectorType from an enumeration into a bit field,
  so that it also can keep track of whether something is a connector,
  expandable connector, or expandable connector element.
- Fixed connector form error message so that the whole connector name is
  printed out.
- Changed OCConnectionGraph.handleOverconstrainedConnections to take a
  Connections record, so it only needs to be created once and so that
  expandable connector can be handled first.

- Removed the list of broken equations from Equation.CONNECT since it
  wasn't used.

### DIFF
--- a/Compiler/NFFrontEnd/NFClass.mo
+++ b/Compiler/NFFrontEnd/NFClass.mo
@@ -469,6 +469,45 @@ uniontype Class
     end match;
   end getType;
 
+  function setType
+    input Type ty;
+    input output Class cls;
+  algorithm
+    () := match cls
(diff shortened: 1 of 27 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2019-02-25_12-52-50](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6464/console).

PR #2961, opened 2019-03-04
[NF] Inner/outer fixes.

perost:
- Fix lookup of inner components by instantiating the class tree of a
  derived class earlier, to make sure that the node inside the
  InstNodeType.BASE_CLASS() of the base class node points to the
  instance created by Inst.instClass and not the shared expanded node.
- Fix adding generated inner components when the class being
  instantiated is a short class definition.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2019-03-04_15-56-50](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6485/console).

PR #2962, opened 2019-03-04
[NF] Hack to fix some array conditions.

perost:
- Handle component and if expression conditions that evaluate to
  arrays (due to array modifiers) by checking if all scalar elements
  are the same value, and in that case use that value.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2019-03-04_17-37-31](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6486/console).

PR #2983, opened 2019-03-18
support getSubtypeOf in NF as well

adrpo:


adrpo:
Bad PR, will open a new one.

PR #2984, opened 2019-03-19
[NF] Function scoping improvements.

perost:
- Use the instance scope instead of the class scope when prefixing
  functions.
- Use the correct scope when instantiating derived function modifiers.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2019-03-19_15-15-40](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6519/console).

PR #2985, opened 2019-03-19
Drop deprecated LAPACK apis

jschueller:
dgelsx, dgegv, dgeqpf are deprecated since lapack 3.1.1 (2007), remove them


perost:
Those functions are still used by the MSL, so we need to support them.

jschueller:
ok, let's see what they say https://github.com/modelica/ModelicaStandardLibrary/pull/2857

PR #2990, opened 2019-03-22
[NF] Improved parameter evaluation analysis.

perost:
- Avoid evaluating parameters that are non-final or depend on non-final
  parameters when Evaluate=true or -d=evaluateAllParameters is used.
- Evaluate parameters that are both fixed and final if their bindings
  are also fixed and final.
- Propagate finalness from modifiers to components, to better exploit
  the above optimization.

### DIFF
--- a/Compiler/NFFrontEnd/NFBinding.mo
+++ b/Compiler/NFFrontEnd/NFBinding.mo
@@ -605,6 +605,24 @@ public
     end match;
   end mapExp;
 
+  function containsExp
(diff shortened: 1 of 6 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2019-03-22_15-24-52](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6529/console).

PR #2992, opened 2019-03-25
Initial implementation of String vars in the Cpp runtime

atrosinenko:
Tested with model:

```modelica
model Test
  output String str;
  output Integer n;
initial algorithm
  str := "123";
algorithm
  when time > 0.1 then
    str := "TEST";
    n := 1;
  end when;
end Test;
```

Before change: cpp code is generated, cannot be compiled to FMU.
After change: FMU is generated and simulated as expected.


atrosinenko:
In fact, this PR is not so "theoretical" even now, since the C runtime has [memory leaks](https://trac.openmodelica.org/OpenModelica/ticket/5196) that sometimes make it unusable, while the Cpp runtime, not unexpectedly, fixes at least some of them.

atrosinenko:
Should this PR be somehow improved?

atrosinenko:
Can this PR be merged or should it be somehow rewritten? On one hand, it may be quite slow implementation. On the other hand, before this PR, it cannot compile at all, so this seems to be improvement, not a regression (unless it is implemented incorrectly in its logic, of course). Sorry to disturb, but this seems to be a blocker for me ("seems" because Cpp runtime may prove still unusable for me after this PR, but I don't know yet).

adrpo:
@AnHeuermann and  @niklwors can you have a quick look at this PR and see if it can be merged in?

adrpo:
I'll start a PR to check if there are no regressions.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2019-04-09_11-25-31](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6664/console).

niklwors:
@adrpo the PR is fine, so it can be merged in the master

AnHeuermann:
Looks good and the provided model works for me.
Maybe we should save (output) string variables in the result file, as there is no output for that at the moment.

lochel:
> Maybe we should save (output) string variables in the result file, as there is no output for that at the moment.

Would be nice, but it would also require a different result file format.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2019-04-09_15-33-34](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6667/console).

atrosinenko:
Thank you!

PR #2993, opened 2019-03-25
Do not run both event and output timers at the same time

sjoelund:
When doing event iteration, we sometimes emit to the result-file. When
doing so, we should not include this time in the time it takes to
perform event iteration.

Fixes https://trac.openmodelica.org/OpenModelica/ticket/5381


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2019-03-25_15-04-00](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6531/console).

PR #2994, opened 2019-03-25
Fix integer packing in C code generation.

perost:
- Pack integers into new arrays instead of packing in place, to avoid
  segfaults when passing integer array literals to external functions.

### DIFF
--- a/Compiler/Template/CodegenCFunctions.tpl
+++ b/Compiler/Template/CodegenCFunctions.tpl
@@ -2197,7 +2197,9 @@ template extFunCallVardecl(SimExtArg arg, Text &varDecls, Text &auxFunction, Boo
   case SIMEXTARG(isInput = true, isArray = true, type_ = ty, cref = c) then
     match expTypeShort(ty)
     case "integer" then
-      'pack_integer_array(&<%contextCref(c,contextFunction,&auxFunction)%>);<%\n%>'
+      let var_name = '<%contextCref(c, contextFunction, &auxFunction)%>'
+      let &varDecls += 'integer_array <%var_name%>_packed;<%\n%>'
+      'pack_alloc_integer_array(&<%var_name%>, &<%var_name%>_packed);<%\n%>'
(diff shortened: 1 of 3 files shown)


OpenModelica-Hudson:
The job failed to compile; for details, see [OpenModelica_TEST_PULL_REQUEST 2019-03-25_15-22-21](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6532/console).

OpenModelica-Hudson:
The job failed to compile; for details, see [OpenModelica_TEST_PULL_REQUEST 2019-03-25_15-31-49](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6533/console).

sjoelund:
What external C function caused a segmentation fault? They are supposed to be passed as const arrays because you are not allowed to modify them in Modelica...

perost:
> What external C function caused a segmentation fault? They are supposed to be passed as const arrays because you are not allowed to modify them in Modelica...

Any external C function that takes an integer array and is called with an array literal. One example is in https://trac.openmodelica.org/OpenModelica/ticket/4904. The NF has previously avoided evaluating arguments of external functions, but with my recent changes that marks more components as final it seems like the backend now evaluates some arguments instead. This causes issues with e.g. Buildings: https://libraries.openmodelica.org/branches/newInst/Buildings_latest/files/Buildings_latest_Buildings.BoundaryConditions.SkyTemperature.Examples.BlackBody.sim

sjoelund:
Oh yeah, the packing modifies it... Better fix would still be to generate a literal of the correct size for the external call in that case.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2019-03-25_17-09-56](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6536/console).

PR #2995, opened 2019-03-25
Fixed ticket:5400 Update the program

adeas31:


perost:
The actual issue seems to have been that `Interactive.transformClassInProgram` forgot to accumulate the classes it had already searched through, so all classes listed before the named class were discarded. Calling updateProgram works I guess, but in a very inefficient way. I fixed in properly in #2996, so this PR should hopefully not be needed anymore.

adeas31:
Yes #2996 is definitely better. We can close this PR.
I also updated the `moveClass` test to check for loaded classes. Use https://github.com/OpenModelica/OpenModelica-testsuite/pull/1146 when you run #2996.

perost:
> I also updated the `moveClass` test to check for loaded classes. Use [OpenModelica/OpenModelica-testsuite#1146](https://github.com/OpenModelica/OpenModelica-testsuite/pull/1146) when you run #2996.

Ok, #2996 is queued up together with OpenModelica/OpenModelica-testsuite#1146 in Hudson, should be ready to process it soon.

PR #3001, opened 2019-03-26
[NF] Improve error checking of when-clauses.

perost:
- Moved the check of equations inside when-equations from Inst to
  Typing, to allow skipping the check if the when is clocked.
- Mark lhs subscripts in when-equations as structural.
- Added check that clocked when doesn't have an elsewhen.
- Added check that non-clocked when doesn't have a clocked elsewhen.
- Made clocked when illegal in algorithm.
- Added new phase VerifyModel that checks for errors that can't be
  detected until after flattening and constant evaluation.
  It currently checks that each branch of a when-equation contains
  the same set of crefs.
- Implemented BaseAvlSet.isEqual.
- Renamed Expression.ClockKind.toString to toDebugString, and
  implemented new toString that doesn't leak implementation details.

### DIFF
--- a/Compiler/NFFrontEnd/NFExpression.mo
+++ b/Compiler/NFFrontEnd/NFExpression.mo
@@ -133,27 +133,45 @@ public
 	    ock := match ick
 	      local
 	        Expression i, ic, r, c, si, sm;
-	      case (INFERRED_CLOCK()) then DAE.INFERRED_CLOCK();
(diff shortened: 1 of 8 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2019-03-26_16-01-18](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6547/console).

PR #3006, opened 2019-03-28
Try even harder to fix Windows builds

sjoelund:


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2019-03-28_09-36-43](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6556/console).

casella:
Lots of [regressions](https://libraries.openmodelica.org/branches/history/newInst/2019-03-27%2018:14:13..2019-03-28%2010:12:55.html) caused by this PR

sjoelund:
Those are just natural fluctuations in performance. They only touched the Windows makefiles so nothing can actually have changed.

casella:
@sjoelund, I wish we could finally give enough juice and time to the testing process so that we avoid having models that run or fail just because the elapsed time is a bit above or below the timeout. This will give us much more consistent output of regression testing.

Can you do that?

adrpo:
The regressions are because of the bad linking with Tables. There are missing symbols which suggests that it needs linking with -lz or so.

adrpo:
We really need different colours for performance!

sjoelund:
Just adding time for the timeout doesn't help because the machine is shared. It will just make a different set of models randomly fail. And increasing the timeout globally makes some older omc versions run for a very long time...

adrpo:
I now changed the colors for performance improvements/regressions so we see it better:
https://github.com/OpenModelica/OpenModelicaLibraryTesting/commit/514c1743f855d390799658cee81f9e0f6c996099

PR #3009, opened 2019-03-28
[NF] Fix when-branch cref set check.

perost:
- Use lists instead of AVL trees for the sets, since the order of
  insertion changes the structure of the trees such that structural
  equality can't be used for the check.
- Remove BaseAvlSet.isEqual, it's no longer needed and might be
  misleading due to only checking structural equality.

### DIFF
--- a/Compiler/NFFrontEnd/NFComponentRef.mo
+++ b/Compiler/NFFrontEnd/NFComponentRef.mo
@@ -568,6 +568,18 @@ public
     end match;
   end isEqual;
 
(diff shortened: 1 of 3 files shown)


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2019-03-28_11-51-35](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6558/console).

PR #3057, opened 2019-04-18
[NF] inStream/actualStream improvements.

perost:
- Fix infinite loop when inStream/actualStream is used in reductions,
  by expanding the reduction argument instead of treating it like an
  array constructor (since reductions does not expand to arrays).
- Added checking of the argument to inStream/actualStream to make sure
  it's a stream variable with parameter subscripts.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2019-04-18_16-39-28](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6689/console).

PR #3060, opened 2019-04-23
[NF] Fix structural analysis of if-condition.

perost:
- Remove marking of conditions in if-equations as structural during the
  instantiation, it can't be done correctly at that point and is already
  done correctly during the typing anyway.


OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2019-04-23_11-28-23](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6692/console).

PR #3064, opened 2019-04-24
Load libraries case insensitive on case sensitive OS

sjoelund:
Libraries where files are stored in the unexpected files will still give
warnings, but OpenModelica will load them regardless.

### DIFF
--- a/Compiler/FrontEnd/ClassLoader.mo
+++ b/Compiler/FrontEnd/ClassLoader.mo
@@ -403,8 +403,12 @@ algorithm
   end if;
   (cl as Absyn.CLASS(name=cname,body=body,info=info))::{} := cs;
   if not stringEqual(cname,pack) then
-    Error.addSourceMessage(Error.LIBRARY_UNEXPECTED_NAME, {pack,cname}, info);
-    fail();
+    if stringEqual(System.tolower(cname), System.tolower(pack)) then
+      Error.addSourceMessage(Error.LIBRARY_UNEXPECTED_NAME_CASE_SENSITIVE, {pack,cname}, info);
+    else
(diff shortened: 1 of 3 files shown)


Compiler/FrontEnd/ClassLoader.mo
```diff
@@ -525,18 +530,32 @@ algorithm
...
     case (CLASSLOAD(str),_,_)
-      equation
-        pd = Autoconf.pathDelimiter;
-        Error.assertionOrAddSourceMessage(System.directoryExists(mp + pd + str) or System.regularFileExists(mp + pd + str + (if encrypted then ".moc" else ".mo")),Error.PACKAGE_ORDER_FILE_NOT_FOUND,{str},info);
+      algorithm
+        pd := Autoconf.pathDelimiter;
+        str2 := str + (if encrypted then ".moc" else ".mo");
+        if not (System.directoryExists(mp + pd + str) or System.regularFileExists(mp + pd + str2)) then
+          try
+            str3 := List.find(System.moFiles(mp), function Util.stringEqCaseInsensitive(str2=System.tolower(str2)));
```
hkiel: The `tolower()` call is not needed for `stringEqCaseInsensitive()` as it does the `tolower()` itself.

OpenModelica-Hudson:
The tests run correctly according to [OpenModelica_TEST_PULL_REQUEST 2019-04-24_09-34-03](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6700/console).

Compiler/FrontEnd/ClassLoader.mo
sjoelund: True...

PR #3079, opened 2019-05-07
Prefer tearing variables with start value for initialization

lochel:
***Related Issues***
[ticket:5458](https://trac.openmodelica.org/OpenModelica/ticket/5458)

***Purpose***
Prefer tearing variables with start value for initialization.

### DIFF
--- a/Compiler/BackEnd/BackendDAEUtil.mo
+++ b/Compiler/BackEnd/BackendDAEUtil.mo
@@ -7521,14 +7521,12 @@ end createGlobalKnownVarsEquations;
  ************************************************/
 
 public function getIndexReductionMethodString
-" function: getIndexReductionMethodString"
   output String strIndexReductionMethod;
 algorithm
   strIndexReductionMethod := Config.getIndexReductionMethod();
 end getIndexReductionMethodString;
 
 public function getIndexReductionMethod
-" function: getIndexReductionMethod"
   input Option<String> ostrIndexReductionMethod;
   output tuple<BackendDAEFunc.StructurallySingularSystemHandlerFunc,String,BackendDAEFunc.stateDeselectionFunc,String> IndexReductionMethod;
 protected
@@ -7545,7 +7543,6 @@ algorithm
 end getIndexReductionMethod;
 
 protected function selectIndexReductionMethod
-" function: selectIndexReductionMethod"
   input String strIndexReductionMethod;
   input list<tuple<Type_a,String,Type_b,String>> inIndexReductionMethods;
   output tuple<Type_a,String,Type_b,String> outIndexReductionMethod;
@@ -7582,14 +7579,12 @@ end selectIndexReductionMethod;
  ************************************************/
 
 public function getMatchingAlgorithmString
-" function: getMatchingAlgorithmString"
   output String strMatchingAlgorithm;
 algorithm
   strMatchingAlgorithm := Config.getMatchingAlgorithm();
 end getMatchingAlgorithmString;
 
 public function getMatchingAlgorithm
-" function: getIndexReductionMethod"
   input Option<String> ostrMatchingAlgorithm;
   output tuple<BackendDAEFunc.matchingAlgorithmFunc,String> matchingAlgorithm;
 protected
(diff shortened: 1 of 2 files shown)


lochel:
@casella @ptaeuber I couldn't remember the discussion on #1668. How should we deal with it?

OpenModelica-Hudson:
The test suite is unstable according to [OpenModelica_TEST_PULL_REQUEST 2019-05-07_14-31-46](https://test.openmodelica.org/hudson//job/OpenModelica_TEST_PULL_REQUEST/6736/).

lochel:
There are two models in the testsuite that give bad results:
* simulation/libraries/3rdParty/ThermoSysPro/ThermoSysPro.Examples.SimpleExamples.TestStodolaTurbine.mos
* simulation/Modelica/inlineFunction/inlineArray1.mos

casella:
> 
> 
> @casella @ptaeuber I couldn't remember the discussion on #1668. How should we deal with it?

As I understand, the discussion there refers to variables that have some non-zero start attribute, no matter how it is defined. 

The requirement here is different: I would try to favour the choice of those variables who have a start attribute _explicitly set in the model_ (or in one of its ancestor) by modifiers, as in the case of the electrical model mentioned in the ticket, see [the source code](https://github.com/modelica/ModelicaStandardLibrary/blob/b7123a80be5897e05e1e27d38d1aaf043b98fdf5/Modelica/Electrical/Analog/Examples/OpAmps.mo#L788). The rationale is that if the modeller adds such a modifier explicitly, it is because he/she wants to influence the convergence, so ignoring it in general is not a good idea, as this specific case clearly demonstrates.

lochel:
Exactly. Most of the test do no longer show the warning about iteration variables without start values. It is only the two tests which produce bad results:

* inlineArray1.mos is not compiling anymore
* ThermoSysPro.Examples.SimpleExamples.TestStodolaTurbine.mos shows now the following warning but simulates fine otherwise:
The Tearing heuristic has chosen variables with annotation attribute 'tearingSelect = avoid'. Use -d=tearingdump and -d=tearingdumpV for more information.

casella:
> Exactly. Most of the test do no longer show the warning about iteration variables without start values.

Well, this seems in fact a very positive outcome, isn't it? 😄 

> It is only the two tests which produce bad results:
> 
>     * inlineArray1.mos is not compiling anymore
I'll have a look at it ASAP.

>     * ThermoSysPro.Examples.SimpleExamples.TestStodolaTurbine.mos shows now the following warning but simulates fine otherwise:
>       The Tearing heuristic has chosen variables with annotation attribute 'tearingSelect = avoid'. 
>       Use -d=tearingdump and -d=tearingdumpV for more information.

This attribute is not coming from the source code, but was added by some OMC backend step. I have currently no idea why OMC would rather avoid using a tearing variable that the modeller gave a start value, but I will check that as well.

casella:
>     * inlineArray1.mos is not compiling anymore

I had a closer look at the BLT this test case. Vector `z` is only function of time, so it can be computed right away. `a[2:5]` is a function of `z`, so it can be computed next. The next strong component has `x[1:5]` and `a[1]` as unknowns, and can be solved by using `a[1]` as tearing variable.

Now, according to the proposal of [ticket:5458](https://trac.openmodelica.org/OpenModelica/ticket/5458), since the whole vector `a` has a local start modifier, all its scalar elements should be preferred in the choice of the tearing variables. (BTW, there is no way to only apply a start value modifier to some elements in an array, unless you introduce additional scalar alias variables, so if we have arrays that's always going to be the case).

However, once the BLT analysis gets to this point, the sub-vector `a[2:5]` has already been matched to other equations, so `a[1]` is the only remaining preferred candidate as tearing variable, as well as the best choice, and the one which older versions of OMC would pick.

 @lochel I'm not sure why this PR should break such a case, maybe the modified algorithm in this PR tries too hard to select _all_ the variables with start modifiers in the array, including the ones that have already been matched, and fails for that reason. No reason to do that, preferred doesn't mean mandatory. Can you please check?

As to the other problematic test case, I guess the root cause of the issue is similar, please check again once the previous case has been dealt with.

